PR #23, opened 2018-09-28
Mem refactor

ttgenius:
modified readwaveforms, calculate_ims to process stations in batches in order to reduce memory usage.
Tested with pytest, hpc


calculate_ims.py
```diff
@@ -430,6 +462,22 @@ def mkdir_output(arg_output, arg_identifier, arg_simple_output):
...
 
 
+def get_steps(input_path, nps, total_stations):
+    """
+    :param input_path: user input file/dir path
+    :param nps: number of processes
+    :param total_stations: total number of stations
+    :return: number of stations per iteration/batch
+    """
+    binary_size = os.stat(input_path).st_size * CORES
```
jasonmotha: Rename CORES and binary_size to their actual representation.

calculate_ims.py
```diff
@@ -430,6 +462,22 @@ def mkdir_output(arg_output, arg_identifier, arg_simple_output):
...
+    nps_total = nps * MEM_PER_CORE
```
jasonmotha: again rename nps_total to what that variable contains.

PR #25, opened 2019-01-15
Allow specification of ouput dir without extra folders

claudio525:


calculate_ims.py
```diff
@@ -445,15 +445,12 @@ def validate_period(parser, arg_period, arg_extended_period, im):
     return period
 
 
-def mkdir_output(arg_output, arg_identifier, arg_simple_output):
+def mkdir_output(output_dir, arg_simple_output):
     """
     create big output dir and sub dir 'stations' inside the big output dir
-    :param arg_output:
-    :param arg_identifier:
     :param arg_simple_output: Boolean. Create 'a subfolder to store single station csvs if True
```
jasonmotha: The comment is backwards. (not you but worth fixing while we are here)

PR #40, opened 2019-04-10
Test calc rrups

ttgenius:


test/test_common_set_up.py
```diff
@@ -1,4 +1,5 @@
 import os
+import pickle
```
jpa198: Doesn't seem to be used?

PR #41, opened 2019-04-11
Geom only

ttgenius:
don't merge yet, just want to see how many test failing


ttgenius:
not relevant any more

PR #43, opened 2019-04-25
Comp change

ttgenius:
comp has been changed from argument parser
pickle data updated

### DIFF
--- a/IM/read_waveform.py
+++ b/IM/read_waveform.py
@@ -209,6 +209,7 @@ def read_one_station_from_bbseries(
         comp = Ellipsis
     else:
         comp = comps
+    print("comps", comps)
     try:
         if wave_type == "a":
             waveform.values = bbseries.acc(
@@ -219,7 +220,10 @@ def read_one_station_from_bbseries(
     except KeyError:
         sys.exit("station name {} does not exist".format(station_name))
     # keep specified comps (remove 3rd column if comps=[0,1]
+    print("befire", waveform.values)
     waveform.values = waveform.values[:, comps]
+    print("after", waveform.values)
+
     return waveform
 
 
--- a/IM/read_waveform.py
+++ b/IM/read_waveform.py
@@ -117,7 +117,6 @@ def read_waveforms(
     bbseis,
     station_names=None,
     comp=Ellipsis,
-    geom_only=False,
     wave_type=None,
     file_type=None,
     units="g",
@@ -140,7 +139,6 @@ def read_waveforms(
         return read_binary_file(
             bbseis,
             comp,
-            geom_only,
             station_names,
             wave_type=wave_type,
             file_type="binary",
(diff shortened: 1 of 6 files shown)


calculate_ims.py
```diff
@@ -60,49 +59,58 @@
...
+    # int_comps = [(1, '000'), (3, 'geom')]
+    sorted_comps = sorted([(EXT_IDX_DICT[c], c) for c in arg_comps])
+    # [1, 3]
+    int_comps = [comp_tuple[0] for comp_tuple in sorted_comps]
+    # ["000", "geom"]
+    str_comps = [comp_tuple[1] for comp_tuple in sorted_comps]
+    if "geom" in arg_comps:
+        int_comps = list(set(int_comps).union({EXT_IDX_DICT["090"], EXT_IDX_DICT["000"]}))  # [0,1,3]
+        int_comps.remove(EXT_IDX_DICT["geom"])  # [0,1]
+        str_comps = ["090", "000", "geom"]
```
jasonmotha: str_comps should be determined after int_comps for both cases.

IM/read_waveform.py
```diff
@@ -215,6 +218,9 @@ def read_one_station_from_bbseries(
             waveform.values = bbseries.vel(station=station_name, comp=comp)
     except KeyError:
         sys.exit("station name {} does not exist".format(station_name))
+    # keep specified comps (remove 3rd column if comps=[0,1]
```
jasonmotha: close bracket. If the parentheses are for an example put e.g. at the front.

calculate_ims.py
```diff
@@ -627,9 +586,10 @@ def main():
...
     parser.add_argument(
         "-c",
-        "--component",
-        type=str,
-        default="ellipsis",
+        "--components",
+        nargs="+",
+        choices=EXT_IDX_DICT.keys(),
+        default=list(EXT_IDX_DICT.keys()),
         help="Please provide the velocity/acc component(s) you want to "
```
jasonmotha: Move the entire help string to the function for tidyness.

calculate_ims.py
```diff
@@ -243,7 +251,7 @@ def compute_measures_multiprocess(
     :param simple_output:
     :return:
     """
-    converted_comp = convert_str_comp(comp)
+    int_comps, str_comps = convert_str_comp(comp)
```
jasonmotha: If int comps aren't used in calculate_ims, we should probably move that logic to where it is needed.

IM/read_waveform.py
```diff
@@ -3,6 +3,8 @@
 
 import numpy as np
 
+from calculate_ims import EXT_IDX_DICT
```
jasonmotha: Can be defined here - to reduce co-dependency. And in calculate_ims EXT_IDX... and IDX_EXT... can be refactored into a component list as it isn't really used anymore.

PR #47, opened 2019-05-15
Advanced IM - Python TCL intergration

jasonmotha:


jasonmotha:
@claudio525 The main thing I'd like reviewed in Advanced_IM_factory. The rest of the changes should be minor. And it's just adding TCL models to the code.

Advanced_IM/advanced_IM_factory.py
```diff
@@ -1,6 +1,75 @@
...
+                    im,
+                ]
+                script.extend(get_filenames_from_components(acc_file, component))
+
+                print(" ".join(script))
+                # subprocess.call(script)
+
+
+def get_filenames_from_components(acc_file, component):
+    return map(lambda x: acc_file[x].name, component)
```
claudio525: I prefer list comprehensions in these scenarios, in my opinion easier to read. Also they are faster (https://stackoverflow.com/questions/1247486/list-comprehension-vs-map, assuming that that is still correct given that the question is 9 years old) when using lambda functions.

But thats just my preference, so happy for it to stay as is.

Advanced_IM/advanced_IM_factory.py
```diff
@@ -1,6 +1,75 @@
...
+
+
+def save_waveform_to_tmp_files(acc_file, accelerations):
+    for component in COMP_DICT.keys():
+        nt = accelerations.shape[0]
+        divisible = nt - nt % VALUES_PER_LINE
+        np.savetxt(
```
claudio525: Two saves?

claudio525:
Looks good, just some minor stuff. 
And maybe some more docstrings, comments in general (I like comments and docstrings haha)

Advanced_IM/advanced_IM_factory.py
jasonmotha: Yeah. because it's saved in columns of 6, so the first save reshapes the data and then the second one prints out the last line. Because I'm passing a file handle it doesn't overwrite the file.

Advanced_IM/advanced_IM_factory.py
jasonmotha: List comperhension is so much better.

PR #50, opened 2019-06-05
Qsw 1236

claudio525:


setup.py
```diff
@@ -0,0 +1,28 @@
+"""
+command: python setup_rspectra.py build_ext --inplace
```
jasonmotha: update comment

setup_rspectra.py
```diff
@@ -1,14 +1,22 @@
 """
```
jasonmotha: Rename file, maybe to setup_local_unit_tests.py

jasonmotha:
And changelog!

jasonmotha:
There needs to be a corresponding change in environments for installing IM_calc?

PR #59, opened 2019-09-15
Rx and Ry

jpa198:


IM_calculation/scripts/calculate_rrups_single.py
```diff
@@ -31,6 +32,10 @@ def write_source_2_site_dists(
         data.append(r_x)
         cols_names.append(SourceToSiteDist.R_x.str_value)
 
+    if r_y is not None:
```
jasonmotha: Let's change this to always write r_x and r_y

PR #62, opened 2019-09-19
Change top/bottom edge calculations to assume they all exist in the f…

jpa198:
…irst/last nstrike locations of each plane of srf points.

Removed unneeded variables from the test data.
Added ability to pass locations with arbitrary additional data after the lon, lat at the start.


jpa198:
> Would we be better off using the SRF header instead of points then?

No, the header/info file don't have enough info in them to reconstitute the corners. NHMs have dipdir, which doesn't seem to be present in either headers or the info file

jpa198:
This should probably try and get the line of best fit through each edge of the plane, instead of just taking the end points, due to the perturbations

PR #65, opened 2019-11-06
temporary fix for conflict between cython and setuptools. referrence:…

shot0829:
… https://github.com/pypa/setuptools/issues/309


jpa198:
My favourite PR so far

PR #66, opened 2019-12-04
Adv im with openseis

shot0829:


IM_calculation/Advanced_IM/advanced_IM_factory.py
```diff
@@ -48,15 +51,14 @@ def compute_ims(accelerations, configuration, adv_im_out_dir):
...
             im_config = config[im]
-            components = ["000", "090", "ver"]
             script = [
                 "python",
                 os.path.join(advanced_im_dir, im_config["script_location"]),
                 "--OpenSees_path",
                 configuration.OpenSees_path,
             ]
             script.extend(
-                [get_acc_filename(f_dir, station_name, x) for x in components]
```
jasonmotha: Ummmmm. Wanna have a closer look at what you've done?

IM_calculation/Advanced_IM/Models/Steel_MF_5Story/run.py
```diff
@@ -46,20 +46,24 @@ def main():
...
+        component_outdir = os.path.join(output_dir, component)
+        #check if folder exist, if not create it (opensees does not create the folder and will crash)
+        if not os.path.isdir(component_outdir):
+            os.makedirs(component_outdir, exist_ok=True)
         script = [
             args.OpenSees_path,
             os.path.join(model_dir, "Run_script.tcl"),
             getattr(args,'comp_'+ component),
-            output_dir,
+            #output_dir,
```
jasonmotha: Remove comment.

IM_calculation/Advanced_IM/Models/SAC_Steel_MF_3Story/run.py
```diff
@@ -0,0 +1,140 @@
+"""
+Python script to run a 3D waveform through Steel_MF_5Story and store the outputs to a txt file
```
jasonmotha: update comment

IM_calculation/Advanced_IM/Models/SAC_Steel_MF_9Story/run.py
```diff
@@ -0,0 +1,140 @@
+"""
+Python script to run a 3D waveform through Steel_MF_5Story and store the outputs to a txt file
```
jasonmotha: comment again

IM_calculation/Advanced_IM/advanced_im_config.yaml
```diff
@@ -6,5 +6,9 @@
 
 Steel_MF_5Story:
   script_location: 'Models/Steel_MF_5Story/run.py'
+SAC_Steel_MF_3Story:
+  script_location: 'Models/SAC_Steel_MF_3Story/run.py'
+SAC_Steel_MF_9Story:
+  script_location: 'Models/SAC_Steel_MF_9Story/run.py'
 SAC_Steel_MFs:
```
jasonmotha: Remove these two lines. This is the two new models now implemented.

IM_calculation/Advanced_IM/advanced_IM_factory.py
shot0829: Dictionary iteration will always stay in the same order, until modified by insertion and such.
ref: [https://docs.python.org/2/library/stdtypes.html#dict.items](https://docs.python.org/2/library/stdtypes.html#dict.items)
re-order the dictionary should solve the 'changed order' issue.

jasonmotha: Can't reorder it either. That's the order that we get the components from BBSeis.

IM_calculation/Advanced_IM/advanced_im_config.yaml
shot0829: but the new models are added in to the repo?
and without these two lines it will not run the new models.

IM_calculation/Advanced_IM/advanced_IM_factory.py
shot0829: since the order of this doesnt really affect the result or usage. i'm marking this as resolved for now.

jasonmotha: Giving that link a read. It says that 'Keys and values are listed in an arbitrary order which is non-random'. 

Aka, even though this time it does actually come out correctly, iterating of a dictionary keys doesn't guarantee that the order will be the same as we inserted into the dictionary.

jasonmotha: Actually, in python 3.7 it is guaranteed to preserve order. But I feel like changing this is just asking for bugs / trouble later. Aka, we don't lose anything being more explicit.

jasonmotha:
@shot0829 Don't forget to black it up.

PR #67, opened 2019-12-19
Sresp

jpa198:


IM_calculation/IM/structural_response.py
```diff
@@ -0,0 +1,436 @@
...
+        moment,
+        storey_moment,
+        shear,
+        storey_shear,
+        load,
+        ground_accel,
+        rel_accel,
+        total_accel,
+    )
+    return disp_peak, moment_peak
```
jasonmotha: Need to find out which measures we actually want to store, From memory it was disp and slope, but need to double check.

IM_calculation/IM/structural_response.py
```diff
@@ -0,0 +1,436 @@
...
+def sa_sd_time(
+    acc, dt, t1_range=None, t1min=1e-06, t1max=5, nt=100, c=0.05, g=9.81, m=1
+):
+    # t1_range: give values instead of calculating them from min, max, nt
+    # t1min: minimum time period to be considered in the design acceleration spectra (seconds).
+    #        0 not chosen to avoid placing 0 in the denominator in later calculations.
+    # t1max: maximum time period to be considered in the design acceleration spectra (seconds).
+    # nt: number of points to be considered in the design acceleration spectra.
+    # c: damping  (c=0.05 represents 5% damping).
+    # g: 9.81 converts acceleration to ms-2;  (if g=1, unit will be in g because the unit of the acceleration record input is in g).
```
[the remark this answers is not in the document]
jpa198: I'm not sure what you mean here. Default behaviour is to convert to ms-2, but if a value of 1 is given then the units are in g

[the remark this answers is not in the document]
jpa198: It should change to be a true/false flag that changes it to g or not, rather than having to give conversion factors yourself?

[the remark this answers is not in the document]
jasonmotha: Yeah. Acceleration should defaultly be in g - but there are a few cases where we have acceleration in cms-2.

So a flag would be perfect rather than having to specify a conversion factor. 

We need to update the comment to say what units are taken in and subsequently what units are outputted.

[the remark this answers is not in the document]
jpa198: I think the comment may be incorrect as the code had the same outputs as Jaes work

IM_calculation/IM/structural_response.py
```diff
@@ -330,7 +335,7 @@ def calculate_structural_response(
     storey_moment = np.cumsum(moment[::-1], axis=0)[::-1]
 
     # calculates total acceleration (floor acceleration) matrix
-    ground_accel = np.tile(g * acc_time_history, (storey + 1, 1))
+    ground_accel = np.tile((1 + g * 8.81) * acc_time_history, (storey + 1, 1))
```
jasonmotha: That seems a bit magicy for me. Can't work out the best way to organise this part though.

Would it be easiest/best it convert the acc_time_history at the start of the function so it's "consistent" all the way through?

IM_calculation/IM/structural_response.py
```diff
@@ -0,0 +1,439 @@
...
+        rel_accel_peak,
+        total_accel_peak,
+    )
+
+
+def get_peak_values(all_values):
+    return np.max(np.abs(all_values), axis=1)
+
+
+def get_structural_response(
```
jasonmotha: Rename this function to Taghavi_Miranda_2005

IM_calculation/IM/structural_response.py
```diff
@@ -0,0 +1,439 @@
...
+    waveform: np.ndarray, dt: float, storey: int, t1: float, alpha: int, c: float
```
jasonmotha: Add a G argument here. (and pass it on upwards)

jasonmotha: And default storey for 10.

IM_calculation/IM/structural_response.py
jasonmotha: Return disp_peak, slope_peak, storey_shear_peak, and total_accel_peak.

PR #70, opened 2020-01-23
Rot d ims

jpa198:


IM_calculation/IM/intensity_measures.py
```diff
@@ -31,15 +31,46 @@ def get_spectral_acceleration(acceleration, period, NT, DT):
...
     # pSA
     if acceleration.ndim != 1:
-        dims = acceleration.shape[1]
+        ts, dims = acceleration.shape
         values = np.zeros((period.size, dims))
+        displacements = np.zeros((period.size, ts * int(round(DT / 0.005)), dims))
         for i in range(dims):
-            values[:, i] = get_spectral_acceleration(acceleration[:, i], period, NT, DT)
-        return values
+            psa, u = get_spectral_acceleration(acceleration[:, i], period, NT, DT)
```
jasonmotha: I thought this was vectorised. Apparently not. Ohhhh It's vectorised across periods though.

IM_calculation/IM/im_calculation.py
```diff
@@ -130,22 +132,30 @@ def compute_measure_single(waveform, ims, comps, im_options, str_comps):
...
-        )
         # store a im type values into a dict {comp: np_array/single float}
         # Geometric is also calculated here
-        result[station_name]["pSA"] = (
-            im_options["pSA"],
-            (array_to_dict(value, str_comps, "pSA", comps)),
+        psa, u = intensity_measures.get_spectral_acceleration_nd(
+            accelerations, im_options["pSA"], waveform_acc.NT, DT
+        )
+        pSA_values = (
```
jasonmotha: A comment of the structure of this would be quite useful.

jasonmotha: It might be a better option to store 'u' as a separate result.

IM_calculation/IM/im_calculation.py
```diff
@@ -250,7 +270,9 @@ def compute_measures_multiprocess(
     :param simple_output:
     :return:
     """
-    str_comps_for_int, str_comps = convert_str_comp(comp)
+    components_to_calculate, components_to_store = constants.Components.get_comps_to_calc_and_store(
+        comp
+    )
 
     bbseries, station_names = get_bbseis(input_path, file_type, station_names)
```
jasonmotha: This doesn't work for txt based waveforms?

IM_calculation/IM/im_calculation.py
```diff
@@ -327,12 +362,16 @@ def write_rows(comps, station, ims, result_dict, big_csv_writer, sub_csv_writer=
     :return: write a single row
     """
     for c in comps:
```
jasonmotha: How does this work if we just create a pandas dataframe?

IM_calculation/IM/read_waveform.py
```diff
@@ -118,7 +120,7 @@ def read_waveforms(
     path,
     bbseis,
     station_names=None,
-    comp=Ellipsis,
+    comp=tuple(Components)[:2],
```
jasonmotha: slicing a tuple?

IM_calculation/scripts/calculate_ims.py
```diff
@@ -109,11 +109,11 @@ def main():
         "-c",
         "--components",
         nargs="+",
-        choices=calc.COMPONENTS,
-        default=calc.COMPONENTS,
+        choices=list(Components.iterate_str_values()),
+        default=list(Components.iterate_str_values())[:4],
```
jasonmotha: Hmmm, we should probably update the default to just geom. I don't see a use case for all 4 at the moment.

IM_calculation/IM/read_waveform.py
claudio525: Is this just a fancy way of getting (Components.c090, Components.c000)? If so maybe just hardcode them, makes it much easier to read I recon.

IM_calculation/IM/read_waveform.py
```diff
@@ -240,7 +237,7 @@ def read_binary_file(
     """
     waveforms = []
     # convert str comps to integer
-    comps = [COMP_DICT[c] for c in comps]
+    # comps = [COMP_DICT[c] for c in comps]
```
claudio525: Is there plans to use this in the future? Otherwise remove it?

IM_calculation/test/test_im_calculation/test_im_calculation.py
```diff
@@ -106,6 +105,9 @@ def test_array_to_dict(self, set_up):
                 os.path.join(root_path, OUTPUT, function + "_value_dict.P"), "rb"
             ) as load_file:
                 expected_value_dict = pickle.load(load_file)
+                for x in list(expected_value_dict.keys())[:]:
```
claudio525: Does the [:] at the end do anything?

IM_calculation/test/test_im_calculation/test_im_calculation.py
```diff
@@ -287,6 +308,21 @@ def test_write_rows(self, set_up):
             sub_csv = io.StringIO()
             sub_csv_writer = csv.writer(sub_csv)
 
+            for station_n in result_dict.keys():
```
claudio525: This logic looks pretty much identical to the one above, worth making a function out of it?

IM_calculation/test/test_im_calculation/test_im_calculation.py
```diff
@@ -338,8 +374,29 @@ def test_write_result(self, set_up):
                 os.path.join(output_folder, calculate_ims.OUTPUT_SUBFOLDER),
                 exist_ok=True,
             )
+            comp = [Components.from_str(c) for c in comp]
```
claudio525: Same again?

IM_calculation/IM/im_calculation.py
jpa198: u isn't a result, it's the spectral displacements which are an intermediate variable (updated for clarity)

IM_calculation/IM/im_calculation.py
jpa198: get_bbseis? It should do, I haven't touched it. The get comps to calc and store should also have the same behaviour as before

IM_calculation/IM/read_waveform.py
jpa198: This was part of a temporary patch while partially refactored towards using the enum

IM_calculation/IM/im_calculation.py
jpa198: Converting it ot use a dataframe sounds like a new task. It should be ok though as you can use to_csv to append to existing csvs, as long as we iterate over stations when saving

jpa198:
@jasonmotha ?

PR #71, opened 2020-02-02
Change IM writing to use a dataframe

jpa198:
Remove deprecated tests
Add extrapolation for Fourier if period is outside available bounds


IM_calculation/IM/im_calculation.py
```diff
@@ -310,10 +309,10 @@ def compute_measures_multiprocess(
             array_params.append(
                 (
                     waveform,
-                    ims,
-                    components_to_store,
+                    sorted(ims),
```
jasonmotha: Should probably sort it asap in the call chain so that way we never have unsorted IMs.

jasonmotha:
And sort your IMs before writing based on qcore/im.py

PR #72, opened 2020-02-14
Update computeFAS.py

lee-robin:
Arguments were the wrong way around in the np.dot.


jasonmotha:
@lee-robin

`fa_smooth = np.dot(fa_spectrum, konno)

E       ValueError: shapes (8193,2) and (8193,8193) not aligned: 2 (dim 1) != 8193 (dim 0)`

Did we need to take a transpose somewhere?

PR #74, opened 2020-04-06
Elastic inelastic ratio test

jpa198:


setup.py
```diff
@@ -37,6 +37,10 @@ def run(self):
         Extension(
             "IM_calculation.IM.rspectra_calculations.rspectra",
             ["IM_calculation/IM/rspectra_calculations/rspectra.pyx"],
+        ),
+        Extension(
```
jasonmotha, line 41: If you use pip does it automatically build the two cython-y bits?

jpa198: Yeah, it should build everything in the list

PR #76, opened 2020-09-16
Made a matrix loader so each thread only has to load the required Kon…

jpa198:
…no matrix once

Fixed bug with RotD saving

This seemed to reduce FAS run time significantly (Bloated out on 20 cores overnight to minutes)


jpa198:
The main purpose of this PR is to prevent a case of The Matrix Reloaded

jasonmotha (approved):

jasonmotha (approved):

PR #77, opened 2020-09-24
Runpy update

shot0829:


IM_calculation/scripts/calculate_ims.py
```diff
@@ -60,7 +73,7 @@ def main():
         help="Please specify the version of the simulation. eg.18p4",
     )
     parser.add_argument(
-        "-m",
+        "-im",
```
jasonmotha, line 76: Does any script use the -m argument?

jpa198 (changes requested):
Lots of comments in common with the 4 run scripts:
A few duplicated functions 
Arg parsing can be centralised
exists_ok=True removes need for isdir checks

IM_calculation/Advanced_IM/Models/SAC_Steel_MF_3Story/run.py
```diff
@@ -0,0 +1,155 @@
+"""
```
jasonmotha, line 1: With there being so many run.py-s now we should do some genericisation of these. Maybe James has idea, he's good at generic stuff.

jasonmotha (changes requested):

IM_calculation/scripts/calculate_ims.py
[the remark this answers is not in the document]
shot0829: thats a error while merging.

shot0829: this was a error while mering master to branch. fixing it.

IM_calculation/Advanced_IM/Models/SAC_Steel_MF_9Story/run.py
```diff
@@ -0,0 +1,175 @@
...
+        help="Where the IM_csv file is written to. Also contains the temporary recorders output",
+    )
+
+    parser.add_argument(
+        "--OpenSees_path",
+        default=DEFAULT_OPEN_SEES_PATH,
+        help="Path to OpenSees binary",
+    )
+
+    args = parser.parse_args()
```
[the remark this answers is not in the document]
jasonmotha: And give an option to send a parser through so that you can add additonal model-specific arguments.

IM_calculation/Advanced_IM/tests/test_runpy.sh
```diff
@@ -0,0 +1,43 @@
+#!/bin/bash
+#out_dir=/nesi/nobackup/nesi00213/RunFolder/ykh22/test_adv_im/v19p5p8_no_gr_SAC_3and9/Runs/3792018/3792018/test_station/tmp/test_ATC12
+if [[ $# -lt 1 ]];then
+    echo "please provide 1. the model name 2. level of magnitude(small, moderate,large) 3. path for output (optional)"
```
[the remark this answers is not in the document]
shot0829: robin said we dont have enough data for large mags, yet.

IM_calculation/IM/im_calculation.py
```diff
@@ -294,14 +330,53 @@ def compute_measures_multiprocess(
...
+        else:
+            # calculate IM for stations in this iteration
+            p.starmap(compute_adv_measure, adv_array_params)
+            # read and agg data into a pandas array
+            # loop through all im_type in advanced_im_config
+            for im_type in advanced_im_config.IM_list:
+                # agg_csv(stations_to_run, output_dir, im_type)
+                df_adv_im[im_type] = df_adv_im[im_type].append(
+                    agg_csv(stations_to_run, output, im_type)
+                )
```
jpa198: Swap the order so you don't need the Not in the conditional

IM_calculation/Advanced_IM/advanced_IM_factory.py
```diff
@@ -0,0 +1,116 @@
...
+            component,
+        )
+
+def agg_csv(stations, im_calc_dir, im_type):
+    # get csv base on station name
+    # quick check of args format
+    if type(im_type) != str:
+        raise TypeError(
+            "im_type should be a string, but get {} instead".format(type(im_type))
+        )
```
jpa198, line 95: Has incorrect type been an issue in the past? Normally I'd just let the path join fail and tell the user to sort it if it's not the right type.
I haven't looked up usages of this function though

IM_calculation/Advanced_IM/advanced_IM_factory.py
```diff
@@ -0,0 +1,116 @@
...
+    # initial a blank dataframe
+    df = pd.DataFrame()
+
+    # loop through all stations
+    for station in stations:
+        # use glob(?) and qcore.sim_struc to get specific station_im.csv
+        # TODO: define this structure into qcore.sim_struct
+        station_im_dir = os.path.join(im_calc_dir, station)
+        im_type_path = os.path.join(station_im_dir, im_type)
+        im_path = os.path.join(im_type_path, im_type + ".csv")
```
jpa198, line 105: This can be a one liner, you can pass multiple parameters to path.join

IM_calculation/Advanced_IM/Models/ATC_Steel_SCBF_12Story/run.py
```diff
@@ -0,0 +1,24 @@
+"""
+Python script to run a 3D waveform through ATC_Steel_SCBF_12Story/ and store the outputs to a txt file
+"""
+
+import os
+
+from IM_calculation.Advanced_IM import advanced_IM_factory
```
jpa198, line 7: Unused import?

jpa198 (changes requested):

IM_calculation/Advanced_IM/advanced_IM_factory.py
shot0829: this can happen if the config wasn't prepared correctly. doesn't really happen often if you are not playing with the config, but there is no drawback to add a check here?

jasonmotha: I'm not a fan of custom exceptions that aren't caught. It just stops being able to trace the error as well. If you need something like that, an assert may be better suited?

IM_calculation/IM/im_calculation.py
```diff
@@ -262,28 +325,35 @@ def compute_measures_multiprocess(
     ) = constants.Components.get_comps_to_calc_and_store(comp)
 
     bbseries, station_names = get_bbseis(input_path, file_type, station_names)
-
     total_stations = len(station_names)
+    # determine the size of each iteration base on num of processers and mem
```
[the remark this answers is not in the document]
shot0829: i dont know how to reduce it even more. simple im and adv im shares the same loop logic, and already calls "compute_measure_single" and "compute_measure_single" on their own.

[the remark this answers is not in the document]
jpa198: I'm in favour of this being split into two functions, 'compute_simple_measures_multiprocess' and 'compute_advanced_measures_multiprocess' or something

IM_calculation/IM/im_calculation.py
```diff
@@ -353,9 +385,10 @@ def compute_measures_multiprocess(
     process=1,
     simple_output=False,
```
jpa198, line 386: Argument should be used or deleted, while you're there

IM_calculation/IM/im_calculation.py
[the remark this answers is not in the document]
jasonmotha: I think making two functions with the advanced IM content will be good enough. They do share loop logic and probably best not to duplicate them. 

However. Currently these do not run alongside.

jasonmotha:
There are a few comments that have not been resolved that have been marked as resolved without comment/change. @shot0829 Can you please go through the ones that won't change and comment the reason why not.

shot0829:
merging master into branch is probably a better idea for now, which is covered by the over PR

PR #88, opened 2021-02-02
fixed path on some tcl where its pointing to wrong location

shot0829:


sungeunbae (approved):

sungeunbae:
@shot0829  Why is the test failing?

sungeunbae (commented):
Please review the test output

shot0829:
> Please review the test output

wierd...it passed last time, and the only thing i added is a comment o.O

jasonmotha (changes requested):
Awaiting test results. (just putting in a review to make pull-reminders happy)

IM_calculation/scripts/calculate_ims.py
```diff
@@ -29,7 +29,7 @@ def load_args():
 
     parser = argparse.ArgumentParser(parents=[parent_parser], add_help=True)
     #
-    #parser = argparse.ArgumentParser()
```
jasonmotha: Remove these lines.

jasonmotha (changes requested):
In general it looks good.

Can you list out any parts of the merge conflict that ended up being hairy?

IM_calculation/scripts/calculate_ims.py
shot0829: there is one more test that fails, i'll remove all those comments after the test are all passed ( keeping them for reference just for now)

shot0829: they should be all removed now.

jasonmotha (approved):

PR #90, opened 2021-02-03
updated logic on reading ascii waveforms

shot0829:


jpa198 (commented):
Wasn't there also a change from floor to round?

shot0829:
> Wasn't there also a change from floor to round?

since we are now forcing DT to be float32, the round change is not required.

IM_calculation/IM/read_waveform.py
```diff
@@ -70,7 +77,7 @@ def read_ascii_header(fid):
     # second line of the header
     header2 = next(fid).split()
     NT = np.int(header2[0])
-    DT = np.float(header2[1])
+    DT = np.float32(header2[1])
```
sungeunbae, line 80: Add a comment here for future reference

sungeunbae (dismissed):
Add some comments

jasonmotha (approved):

jpa198 (approved):

PR #92, opened 2021-02-23
Iesdr

sungeunbae:
Added IESDR integration to the workflow. It uses the same periods as pSAs and imports Python implementation (not Cython version).
Other parameters (z, alpha, dy) are kept as constants. This version produces the consistent results as Vahid's test script.


IM_calculation/IM/im_calculation.py
```diff
@@ -306,6 +309,37 @@ def calculate_pSAs(
...
+    station_name,
+    waveform_acc,
+    im,
+    comps_to_store,
+    comps_to_calculate,
+):
+    # for the time being, following 3 parameters are fixed
+    z = 0.05  # damping ratio
+    alpha = 0.05  # strain hardening ratios
+    dy = 0.025  # strain hardening ratios
```
jpa198, line 326: These should be function parameters with the given values as defaults, so that we can use different values later on if needed

jpa198 (changes requested):
Please remove the zip file for Vahid
Otherwise looks good

IM_calculation/IM/im_calculation.py
sungeunbae: Good point. Will update shortly

jasonmotha (approved):

jasonmotha (approved):

jpa198 (approved):

PR #95, opened 2021-03-02
Runpy update

shot0829:


IM_calculation/IM/im_calculation.py
```diff
@@ -445,24 +508,58 @@ def compute_measures_multiprocess(
                     sorted(components_to_calculate, key=lambda x: x.value),
                 )
             )
-        all_results.extend(p.starmap(compute_measure_single, array_params))
-
-    all_result_dict = ChainMap(*all_results)
-    output_path = get_result_filepath(output, identifier, ".csv")
+            adv_array_params.append((waveform, advanced_im_config, output))
+        # only run simply im if and only if adv_im not going to run
+        if advanced_im_config.IM_list:
```
jasonmotha, line 513: As discussed in the last PR. Put this Adv IM block into a separate function.

IM_calculation/Advanced_IM/runlibs_2d.py
```diff
@@ -0,0 +1,189 @@
...
+            os.makedirs(component_outdir, exist_ok=True)
+        script = [
+            args.OpenSees_path,
+            run_script,
+            getattr(args, "comp_" + component),
+            component_outdir,
+        ]
+
+        print(" ".join(script))
+        subprocess.call(script)
```
jpa198, line 59: Should generally use run instead of call going forward

IM_calculation/scripts/calculate_ims.py
```diff
@@ -163,19 +191,42 @@ def main():
...
+    advanced_im_config = advanced_IM_factory.advanced_im_config(
+        args.advanced_ims, args.advanced_im_config, args.OpenSees_path
+    )
+
     # Create output dir
     utils.setup_dir(args.output_path)
     if not args.simple_output:
         utils.setup_dir(os.path.join(args.output_path, calc.OUTPUT_SUBFOLDER))
 
+    if args.observed != None:
```
jasonmotha, line 203: I'm not convinced this is the best way to reduce the station's calculated. I guess this is a sandboxed argument and it would work.

jasonmotha (changes requested):

IM_calculation/Advanced_IM/Models/Spear_3D_Model/run.py
```diff
@@ -0,0 +1,311 @@
...
+
+def calculate_norm(im_name, output_dir, print_header=True):
+    # get 000 and 090 dir
+    dir_000 = os.path.join(output_dir, "000")
+    dir_090 = os.path.join(output_dir, "090")
+    component = "norm"
+
+    im_csv_fname = os.path.join(output_dir, im_name + "_" + component + ".csv")
+
+    df_combined = {}
```
jpa198, line 112: Unused?

IM_calculation/Advanced_IM/Models/Spear_3D_Model/run.py
```diff
@@ -0,0 +1,311 @@
...
+    cols.sort()
+
+    # test if file exist, if exist, no header
+    if os.path.isfile(im_csv_fname):
+        print_header = False
+
+    result_df.to_csv(im_csv_fname, mode="a", header=print_header, columns=cols)
+
+
+def calculate_geom(output_dir, im_name):
```
jpa198, line 169: Add a docstring to say this func takes two files in and generates the geom from them?

IM_calculation/Advanced_IM/Models/Taghavi_Miranda_2005/run.py
```diff
@@ -0,0 +1,93 @@
...
+    parser = argparse.ArgumentParser()
+    parser.add_argument(
+        "comp_000", help="filepath to a station's 000 waveform ascii file"
+    )
+    parser.add_argument(
+        "comp_090", help="filepath to a station's 090 waveform ascii file"
+    )
+    parser.add_argument(
+        "comp_ver", help="filepath to a station's ver waveform ascii file"
+    )
```
jpa198, line 35: Unused?

IM_calculation/Advanced_IM/Models/Spear_3D_Model/run.py
```diff
@@ -0,0 +1,311 @@
+import argparse
+import glob
+import os
+import pandas as pd
+import subprocess
+
+import numpy as np
+
+
```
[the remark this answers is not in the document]
shot0829: the issue is that 3D models perform differently, compare to 2D.
and there is only one 3D models at the moment so i wasnt sure if i should have a lib.
But if you prefer, i can make a runlibs_3D for it.

IM_calculation/Advanced_IM/advanced_IM_factory.py
```diff
@@ -0,0 +1,115 @@
...
+advanced_im_config = namedtuple(
+    "advanced_im_config", ["IM_list", "config_file", "OpenSees_path"]
+)
+COMP_DICT = {"090": 0, "000": 1, "ver": 2}
+
+
+def get_config(config_file=CONFIG_FILE_NAME):
+    with open(config_file) as cf:
+        config = yaml.safe_load(cf)
+    return config
```
jpa198, line 24: We already have a yaml loader in qcore?

IM_calculation/Advanced_IM/Models/Spear_3D_Model/run.py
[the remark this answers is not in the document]
jasonmotha: Ahhh I see. Makes sense - we can create a 3D lib when we add the next 3D model. I think it can at least use the argparse from it. It may make sense to have 2D and 3D content in run_libs and have 2D/3D specific functions.

IM_calculation/IM/im_calculation.py
```diff
@@ -37,6 +40,33 @@
 MEM_FACTOR = 4
 
 
+def add_if_not_exist(component_list, component):
+    if not component in component_list:
+        component_list.append(component)
+    return component_list
+
+
+def convert_str_comp(arg_comps):
```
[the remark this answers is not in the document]
shot0829: now you mentioned it, i remember when and why it was removed. but for some reason the testing still testing it...?
i'll do a update.

IM_calculation/Advanced_IM/advanced_IM_factory.py
```diff
@@ -0,0 +1,115 @@
...
+            ]
+            # waveform component sequence
+            comp_list = ["000", "090", "ver"]
+            script.extend([get_acc_filename(f_dir, station_name, x) for x in comp_list])
+            script.extend([out_dir])
+
+            script.extend(["--OpenSees_path", f"{configuration.OpenSees_path}"])
+
+            print(" ".join(script))
+            subprocess.call(script)
```
jpa198, line 64: Another use of call

jpa198 (changes requested):

jpa198:
Now to see how many of my comments were solved while I was reviewing...

IM_calculation/Advanced_IM/Models/Spear_3D_Model/run.py
```diff
@@ -0,0 +1,311 @@
...
+    parser = argparse.ArgumentParser()
+    parser.add_argument(
+        "comp_000", help="filepath to a station's 000 waveform ascii file"
+    )
+    parser.add_argument(
+        "comp_090", help="filepath to a station's 090 waveform ascii file"
+    )
+    parser.add_argument(
+        "comp_ver", help="filepath to a station's ver waveform ascii file"
+    )
```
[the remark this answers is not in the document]
jasonmotha: Because ADV im will pass ver component in regardless when running this run_py

IM_calculation/Advanced_IM/advanced_IM_factory.py
```diff
@@ -0,0 +1,115 @@
...
+
+def save_waveform_to_tmp_files(tmp_folder, accelerations, station_name):
+    """
+    Writes to the 3 files containing values for all components
+    :param acc_file: Dict containing file handles for each component specified
+    :param accelerations: Acceleration array, 1 column for each component. Ordering is specified in COMP_DICT
+    :return: None
+    """
+    print(f"accelerations {accelerations}")
+    # TODO: Fix bug when differing components are specified in calculate_ims
```
jasonmotha, line 80: I'm not sure that's the best idea. It adds a lot of complexity to the models especially ones that need multiple components. It'll be quite a big change.

IM_calculation/Advanced_IM/Models/Spear_3D_Model/run.py
```diff
@@ -0,0 +1,311 @@
...
+
+def main():
+
+    parser = argparse.ArgumentParser()
+    parser.add_argument(
+        "comp_000", help="filepath to a station's 000 waveform ascii file"
+    )
+    parser.add_argument(
+        "comp_090", help="filepath to a station's 090 waveform ascii file"
+    )
```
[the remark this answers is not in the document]
shot0829: having them separate arguments provide flexibility. the script/wrapper above this handle the naming things.
this will allow us to run scripts when waveforms at extracted into tmp folders (which we currently already does) and it will not break if waveforms are in different tmp.

IM_calculation/Advanced_IM/Models/Spear_3D_Model/run.py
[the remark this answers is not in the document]
shot0829: for this particular 3d model its not being used.
but the 3 component is like a standard input you should have anyways.
and if i'm making a runlibs_3d, as mentioned above with @jasonmotha comments, it have to be generic and cover most cases?

IM_calculation/scripts/calculate_ims.py
```diff
@@ -10,14 +10,27 @@
...
+    parent_parser.add_argument(
+        "--advanced_im_config",
+        default=advanced_IM_factory.CONFIG_FILE_NAME,
+        help="Path to the advanced IM_config file",
+    )
+
+    parent_args = parent_parser.parse_known_args()
+
+    parser = argparse.ArgumentParser(parents=[parent_parser], add_help=True)
+
```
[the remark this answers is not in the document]
jasonmotha: Yes it's way cool.

It needs a separate section because it populates options based on the contents of the config.

IM_calculation/Advanced_IM/Models/Spear_3D_Model/run.py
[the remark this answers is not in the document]
jasonmotha: I don't like assumptions. My thoughts are the current way is better, may be more verbose but this args are behind the scenes anyway.

IM_calculation/Advanced_IM/advanced_IM_factory.py
shot0829: i agree, this line was debugging line i added a while ago and forgot. 
And this is one of the reason that while ver are not being used as comp for some models, it is always required in the argparser @jpa198 .

IM_calculation/Advanced_IM/runlibs_2d.py
jpa198: https://docs.python.org/3/library/subprocess.html#older-high-level-api

IM_calculation/scripts/calculate_ims.py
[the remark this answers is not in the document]
shot0829: this was done before i started working on this branch.
but the design decision is mostly to be that since adv_im and base_im are distinctively different and do not interfere with each other, and if the base_im start become more complicated, having advanced_im_config as a parent parser means it will not be affected and alway be included if required.
@jasonmotha should have more idea about it, as he wrote the 1st version of this code.

IM_calculation/Advanced_IM/advanced_IM_factory.py
shot0829: updated to use subprocess.run()

IM_calculation/Advanced_IM/Models/Spear_3D_Model/run.py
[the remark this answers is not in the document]
jpa198: I'm generally in favour of separate 2d, 3d and common libs.
But if we only have one 3d model atm, then having 3d funcs with the model is fine, they can be moved out later as required

IM_calculation/scripts/calculate_ims.py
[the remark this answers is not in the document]
jpa198: Ah, I see how it works.
I'm not too keen on the way it works and would rather see adv_im be a sub command but won't argue to change it

sungeunbae (dismissed):

sungeunbae:
Oh - I applied a filetype filter, and only chose .py files which only displayed 3 run.pys - it turns out the filter is not good, filtering out many other .py files!

jasonmotha:
I did the filetype filter too. It showed all of them but I had to load the files though.

IM_calculation/IM/im_calculation.py
```diff
@@ -376,11 +376,7 @@ def get_bbseis(input_path, file_type, selected_stations):
             station_names = bb_stations
         else:
             # making sure selected stations are in bbseis
-            station_names = [
-                station_name
-                for station_name in selected_stations
-                if station_name in bb_stations
-            ]
+            station_names = selected_station.intersection(bb_stations)
```
jasonmotha, line 379: Isn't selected_station a list not a set?

jasonmotha (changes requested):

IM_calculation/IM/im_calculation.py
shot0829: fixed

IM_calculation/IM/im_calculation.py
shot0829: i've greatly trim the code and moved the logic to adv_im_factory.

IM_calculation/Advanced_IM/Models/Spear_3D_Model/run.py
```diff
@@ -0,0 +1,288 @@
...
+    return im_records_list
+
+
+def calculate_norm(im_name, output_dir, print_header=True):
+    # get 000 and 090 dir
+    dir_000 = os.path.join(output_dir, "000")
+    dir_090 = os.path.join(output_dir, "090")
+    component = "norm"
+
+    im_csv_fname = os.path.join(output_dir, im_name + "_" + component + ".csv")
```
jpa198, line 86: Format string instead of concatenation

IM_calculation/IM/im_calculation.py
```diff
@@ -119,20 +118,21 @@ def compute_adv_measure(waveform, advanced_im_config, output_dir):
...
-        if advanced_im_config.IM_list is not None:
-            waveform_acc = waveform[0]
-            station_name = waveform_acc.station_name
-            adv_im_out_dir = os.path.join(output_dir, station_name)
+    
+    if advanced_im_config.IM_list is not None:
+        waveform_acc = waveform[0]
+        station_name = waveform_acc.station_name
+        adv_im_out_dir = os.path.join(output_dir, station_name)
+        try:
```
jasonmotha, line 126: Why did you more the try? I think it was there for a reason but I can't remember.

IM_calculation/scripts/calculate_ims.py
jasonmotha: I've been thinking about it and it's not inline with keeping structure / complexity out of calculation scripts. 

If you need that you should write a separate script that does this and then pass the output as an argument into -s (--stations)

jasonmotha:
Making good progress on this! It's really good to see.

IM_calculation/IM/im_calculation.py
jpa198: Is there a case IM_list is not None, but still throws an error? If not can remove the try

jpa198: @jasonmotha move or remove?

jasonmotha: I mean the try was meant to be checking that IM_list existed under advanced_im_config. (that's at least the error that's printed out)

shot0829: in theory, it shouldn't, if the config is passed all the way from adv_im_factory.
but if for some reason this function is called in non-typical way, it might.

jpa198: Oh, so the start of the function should have something like this?
try: if advanced_im_config.IM_list is None: print("Nothing to do"); return
except: print("Broken"); return
else: do the thing

jasonmotha: Kinda. I was suggesting like Jonney's current change.

shot0829: it should break the whole script, since which ever calling this function should have a config.
i've updated it to use if else, its probably easier that way.

IM_calculation/scripts/calculate_ims.py
shot0829: the only cons of doing so it that the execute command in slurm/pbs/etc can be extremely long.
(lets say we have a matching 80 stations)

jasonmotha: I don't personally see that as a big deal. Previously we've ignored the station matching because IM_calc didn't use much resources but can definitely see this being an issue for Adv_IM.

shot0829: i'll remove this from calulate_im for now.
but if we starting to have issue with the length of sysargv, we can always change this to read a tmp/station_list_file thats generated on the fly.

IM_calculation/IM/im_calculation.py
```diff
@@ -445,24 +509,58 @@ def compute_measures_multiprocess(
...
+            p.starmap(compute_adv_measure, adv_array_params)
+            # read and agg data into a pandas array
+            # loop through all im_type in advanced_im_config
+            for im_type in advanced_im_config.IM_list:
+                df_adv_im[im_type] = df_adv_im[im_type].append(
+                    advanced_IM_factory.agg_csv(stations_to_run, output, im_type)
+                )
+        else:
+            result_list = p.starmap(compute_measure_single, array_params)
+            all_result_dict = ChainMap(*result_list)
```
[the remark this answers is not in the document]
shot0829: yeah, that was a mistake, the newest commit on trimming the code had fix this.

IM_calculation/Advanced_IM/advanced_IM_factory.py
```diff
@@ -148,6 +148,8 @@ def agg_csv(advanced_im_config, stations, im_calc_dir):
 
 def natural_key(string_):
     """
-    using regex to get a sequence of numbers in a string and turn thme into int for sorting purpose
+    using regex to get a sequence of numbers in a string and turn them into int for sorting purpose
+    example strings: test_model_v20p1p1  test_model2_v20p1p11
+    keys: [test_model_v, 20, p, 1, p , 1],  [test_model2_v, 20, p, 1, p, 11]
```
jasonmotha, line 153: put quotes on here to indicate which are strings and which are ints.

Is the second example incorrect ```test_model2_v```

shot0829: haha, yeah, i noticed it when i was reading it this morning. updated.

IM_calculation/Advanced_IM/Models/Taghavi_Miranda_2005/run.py
```diff
@@ -0,0 +1,74 @@
...
+                            i
+                        ].storey_shear_peak
+                        results[component][f"{im_name}{i}_total_accel_peak"] = df.iloc[
+                            i
+                        ].total_accel_peak
+
+    im_csv_fname = os.path.join(output_dir, "Taghavi_Miranda_2005.csv")
+    df = pd.DataFrame.from_dict(results, orient="index")
+    df.index.name = "component"
+    geom = pd.Series(np.sqrt(df.loc["000"] * df.loc["090"]), name="geom")
```
jasonmotha, line 67: maybe use the function to calc geom?

IM_calculation/Advanced_IM/advanced_IM_factory.py
```diff
@@ -0,0 +1,155 @@
...
+    assert (
+        type(im_type) == str
+    ), "im_type should be a string, but get {} instead".format(type(im_type))
+    # initial a blank dataframe
+    df = pd.DataFrame()
+
+    # loop through all stations
+    for station in stations:
+        # get csv base on station name
+        # use glob(?) and qcore.sim_struc to get specific station_im.csv
```
jasonmotha, line 105: sim_struct

IM_calculation/Advanced_IM/advanced_IM_factory.py
```diff
@@ -0,0 +1,155 @@
...
+        # TODO: define this structure into qcore.sim_struct
+        im_path = os.path.join(im_calc_dir, station, im_type, im_type + ".csv")
```
jasonmotha, line 107: f_string

IM_calculation/Advanced_IM/advanced_IM_factory.py
```diff
@@ -0,0 +1,155 @@
...
+        adv_im_df_dict[im_type] = adv_im_df_dict[im_type].append(
+            read_csv(stations, im_calc_dir, im_type)
+        )
+        # do a natural sort on the column names
+        adv_im_df_dict[im_type] = adv_im_df_dict[im_type][
+            list(adv_im_df_dict[im_type].columns[:2])
+            + sorted(adv_im_df_dict[im_type].columns[2:], key=natural_key)
+        ]
+        # check if file exist already, if exist header=False
+        adv_im_out = os.path.join(im_calc_dir, im_type + ".csv")
```
jasonmotha, line 138: fstring

IM_calculation/Advanced_IM/runlibs_2d.py
```diff
@@ -0,0 +1,187 @@
...
+
+        # get base name
+        # env_dir = os.path.dirname(im_recorder)
+        # env_name = os.path.basename(env_dir).split('_')[-1]
+        sub_im_type = sub_im_name.split("_")[0]
+        sub_im_gravity_dir = os.path.join(component_outdir, "gravity_" + sub_im_type)
+        sub_im_gravity_recorder = os.path.join(
+            sub_im_gravity_dir, "gr_" + os.path.basename(im_recorder)
+        )
+        # find corrosponding gravity file
```
jasonmotha, line 148: corresponding*

IM_calculation/Advanced_IM/tests/test_runpy.sh
```diff
@@ -0,0 +1,44 @@
+#!/bin/bash
+#out_dir=/nesi/nobackup/nesi00213/RunFolder/ykh22/test_adv_im/v19p5p8_no_gr_SAC_3and9/Runs/3792018/3792018/test_station/tmp/test_ATC12
```
jasonmotha, line 2: Remove commented lines? (and at end of file?)

IM_calculation/IM/im_calculation.py
```diff
@@ -445,23 +474,31 @@ def compute_measures_multiprocess(
...
-
-    results_dataframe = pd.DataFrame.from_dict(all_result_dict, orient="index")
-    results_dataframe.index = pd.MultiIndex.from_tuples(
-        results_dataframe.index, names=["station", "component"]
-    )
-    results_dataframe.sort_values(["station", "component"], inplace=True)
-    results_dataframe = order_im_cols_df(results_dataframe)
+            adv_array_params.append((waveform, advanced_im_config, output))
+        # only run basic im if and only if adv_im not going to run
+        if advanced_im_config.IM_list:
```
jasonmotha, line 479: We're checking IM_list before we get to checking if advanced_im_config is NULL. Can we move the check in compute_adv_measures_ here?

IM_calculation/scripts/calculate_ims.py
```diff
@@ -138,6 +151,22 @@ def load_args():
...
+        "-a",
+        "--advanced_ims",
+        nargs="+",
+        choices=advanced_IM_factory.get_im_list(parent_args[0].advanced_im_config),
+        help="Provides the list of Advanced IMs to be calculated",
+    )
+    parser.add_argument(
+        "--OpenSees_path", default="OpenSees", help="Path to OpenSees binary"
+    )
+    parser.add_argument(
```
jasonmotha, line 164: If block has been removed, can remove argument too.

jasonmotha (changes requested):
Just a few (well maybe a lot) minor things.

IM_calculation/Advanced_IM/Models/Taghavi_Miranda_2005/run.py
```diff
@@ -0,0 +1,74 @@
...
+ALPHA = [0, 1.5, 5, 15, 30]
+C = [0.02, 0.05, 0.1]
+T = [0.1, 0.4, 0.7, 1.5, 3.0]
+STORIES = 10
+
+
+def main(comp_000, comp_090, output_dir):
+
+    if not os.path.exists(output_dir):
+        os.makedirs(output_dir)
```
jpa198, line 27: exist_ok=True?

IM_calculation/Advanced_IM/runlibs_2d.py
```diff
@@ -0,0 +1,187 @@
...
+
+    # skip creating csv if any component has zero success count
+    if model_converged:
+
+        for component in ["000", "090"]:
+            component_outdir = os.path.join(output_dir, component)
+            # aggregate
+            create_im_csv(output_dir, im_name, component, component_outdir)
+
+            im_csv_fname = os.path.join(output_dir, im_name + ".csv")
```
jpa198, line 71: fstring

IM_calculation/Advanced_IM/runlibs_2d.py
```diff
@@ -0,0 +1,187 @@
...
+            calculate_geom(im_csv_fname)
+    else:
+        station_name = os.path.basename(args.comp_000).split(".")[0]
+        im_csv_failed_name = os.path.join(output_dir, im_name + "_failed" + ".csv")
```
jpa198, line 75: fstring

IM_calculation/IM/im_calculation.py
```diff
@@ -110,6 +110,28 @@ def calculate_rotd(
...
+        waveform_acc = waveform[0]
+        station_name = waveform_acc.station_name
+        adv_im_out_dir = os.path.join(output_dir, station_name)
+        advanced_IM_factory.compute_ims(
+            waveform_acc, advanced_im_config, adv_im_out_dir
+        )
+    else:
+        raise AttributeError(
+            f"cannot access advanced_im_config doesnt contains IM_list {advanced_im_config}"
+        )
```
jpa198, line 132: Could use de morgans law to swap the logic, so you're testing for failure and raising, then can deindent the actual body?
Non functional change, just aesthetic

IM_calculation/IM/im_calculation.py
```diff
@@ -413,28 +435,35 @@ def compute_measures_multiprocess(
...
     total_stations = len(station_names)
+    # determine the size of each iteration base on num of processers and mem
     steps = get_steps(
         input_path, process, total_stations, "FAS" in ims and bbseries.nt > 32768
     )
 
     all_results = []
     p = Pool(process)
 
+    if advanced_im_config.IM_list:
```
jpa198, line 447: As per Jasons comment below RE checking IM_LIST

IM_calculation/IM/im_calculation.py
jasonmotha: Agreed with this change. The less indented blocks the better. Especially for terminating conditions.

IM_calculation/IM/im_calculation.py
```diff
@@ -413,28 +435,35 @@ def compute_measures_multiprocess(
...
     bbseries, station_names = get_bbseis(input_path, file_type, station_names)
-
     total_stations = len(station_names)
+    # determine the size of each iteration base on num of processers and mem
     steps = get_steps(
         input_path, process, total_stations, "FAS" in ims and bbseries.nt > 32768
     )
 
     all_results = []
     p = Pool(process)
```
jpa198, line 445: This should be moved to just before the while loop and used with a context manager
Claudio will be more than happy to explain why :P

IM_calculation/IM/im_calculation.py
```diff
@@ -413,28 +435,35 @@ def compute_measures_multiprocess(
...
 
+    if advanced_im_config.IM_list:
+        df_adv_im = {im: pd.DataFrame() for im in advanced_im_config.IM_list}
+
     i = 0
     while i < total_stations:
```
jpa198, line 451: Not requiring a change, but why is this a while loop instead of "for i in range(0, total_stations, steps)" loop?

jasonmotha: Because our code scrutiny wasn't as high as it is now when this code was written.

shot0829: i'm not sure whoever wrote the first version of this, but i personally like while better than for. i think its easier to follow the logic and more flexible ( instead of set sequence, you can do more complex manipulation)

jasonmotha: Imo the only time you should write whiles in python is when you don't know how many iteration steps you are going to take.

IM_calculation/Advanced_IM/Models/Taghavi_Miranda_2005/run.py
shot0829: the function from runlib? that function includes reading csv (output from opensees) and dumping as well.
the function in runlib has more than one step, and makes more sense to be a function; the case we have here is a simple sqrt command, i'm not sure if creating a new function is a good idea?

IM_calculation/IM/im_calculation.py
jpa198: :+1:  to Jason, I'm in favour of this being a for loop *eventually*. Not for this PR though

jasonmotha: Imo the only time you should write whiles in python is when you don't know how many iteration steps you are going to take. Aka iterating until a certain condition is met. There's a lot of potential for bugs with while loops.

IM_calculation/Advanced_IM/Models/Taghavi_Miranda_2005/run.py
jasonmotha: That makes sense. Probably best to use this one though. So that way we aren't defining geom calculation (even though it's simple) in more than one place.

https://github.com/ucgmsim/IM_calculation/blob/cb29fc5faa69ff95444a255a2b75c18e7186656e/IM_calculation/IM/intensity_measures.py#L147

IM_calculation/IM/im_calculation.py
shot0829: What's the toughs behind this? i'm not against it, but doing so means that the function itself is depending on the which even parents script calls it.

IM_calculation/IM/im_calculation.py
jpa198: Does this function even need integrity checks? It should raise exceptions if the data isn't good., it's not like we're putting a default value in.
I think the onus should be on calling functions to ensure they're putting in good information

IM_calculation/scripts/calculate_ims.py
```diff
@@ -163,19 +191,30 @@ def main():
     if "FAS" in im:
         im_options["FAS"] = calc.validate_fas_frequency(args.fas_frequency)
 
+    advanced_im_config = advanced_IM_factory.advanced_im_config(
+        args.advanced_ims, args.advanced_im_config, args.OpenSees_path
+    )
+
```
jpa198, line 197: Move this down to the 
if args.adv_ims != None
And add an adv_im_cfg=None to the else condition

IM_calculation/scripts/calculate_ims.py
```diff
@@ -163,19 +191,30 @@ def main():
...
     # Create output dir
     utils.setup_dir(args.output_path)
     if not args.simple_output:
         utils.setup_dir(os.path.join(args.output_path, calc.OUTPUT_SUBFOLDER))
 
+    # TODO: this may need to be updated to read file if the length of list becomes an issue
+    station_names = args.station_names
+    if args.advanced_ims != None:
```
jpa198, line 205: is not None

IM_calculation/Advanced_IM/advanced_IM_factory.py
```diff
@@ -102,9 +102,9 @@ def read_csv(stations, im_calc_dir, im_type):
     # loop through all stations
     for station in stations:
         # get csv base on station name
-        # use glob(?) and qcore.sim_struc to get specific station_im.csv
-        # TODO: define this structure into qcore.sim_struct
-        im_path = os.path.join(im_calc_dir, station, im_type, im_type + ".csv")
+        # use glob(?) and qcore.simulation_structure to get specific station_im.csv
+        # TODO: define this structure into qcore.simulation_structure
+        im_path = os.path.join(im_calc_dir, station, im_type, "{im_type}.csv")
```
jasonmotha, line 107: You need the f lol

shot0829: oops

IM_calculation/scripts/calculate_ims.py
shot0829: done

IM_calculation/Advanced_IM/advanced_IM_factory.py
shot0829: fixed

IM_calculation/IM/im_calculation.py
shot0829: updated. logic moved one level above

IM_calculation/Advanced_IM/tests/test_runpy.sh
shot0829: done

IM_calculation/Advanced_IM/runlibs_2d.py
shot0829: haha, fixed

IM_calculation/Advanced_IM/advanced_IM_factory.py
shot0829: updated

IM_calculation/Advanced_IM/Models/Taghavi_Miranda_2005/run.py
jpa198: Not actually resolved

IM_calculation/Advanced_IM/Models/Taghavi_Miranda_2005/run.py
jpa198: Yeah, use the get_geom function.

One day we might change geom to be exp((log(d1)+log(d2))/2)

shot0829: the latest commit should have updated the code to use get_geom from IM_calculation.IM.intensity_measures

IM_calculation/Advanced_IM/Models/Taghavi_Miranda_2005/run.py
shot0829: the newest commit resolved this? o.O?
were you looking at a older commit?

IM_calculation/Advanced_IM/Models/Taghavi_Miranda_2005/run.py
shot0829: @jasonmotha since @jpa198 mentioned this potential change, should the geom calculation in OpenSees's 2D model be updated to have this as well(using get_geom)? or should they stay independent?

IM_calculation/IM/im_calculation.py
```diff
@@ -110,6 +110,28 @@ def calculate_rotd(
...
+def compute_adv_measure(waveform, advanced_im_config, output_dir):
+    """
+    Wrapper function to call advanced IM workflow
+    :param waveform: Tuple of waveform objects (first is acc, second is vel)
+    :param advanced_im_config: advanced_im_config Named Tuple containing list of IMs, config file and path to OpenSeeS
+    :param output_dir: Directory where output folders are contained. Structure is /path/to/output_dir/station/im_name
+    :return:
+    """
+
+    if (advanced_im_config is not None) and (advanced_im_config.IM_list is not None):
```
jasonmotha: I'm not sure brackets increase readability. Thoughts?

jpa198: I like that it breaks up the line into "if A and B", and that you can easily distinguish what A and B are. 
Without brackets is probably ok, but I'd rather leave it as is

IM_calculation/Advanced_IM/Models/Taghavi_Miranda_2005/run.py
jasonmotha: Yes. 

```        line = np.sqrt(ims.loc["090"] * ims.loc["000"])````
        
        should be updated to use the function.

jpa198: Please note: The alternate formula is mathematically correct, but is dead slow compared to the current sqrt method.
All geom calculations should still point at the one function though

jasonmotha (approved):

jpa198 (approved):

jpa198:
Do we want to get a third opinion or just put it in?

jasonmotha:
I'm happy to merge this as is.

PR #96, opened 2021-03-04
IESDR2: Ensure numerical stability by adjusting the time step

sungeunbae:
In response to[ Vahid's feedback](https://uceqeng.slack.com/archives/C3SGQHXT8/p1614548240059100?thread_ts=1614037026.009400&cid=C3SGQHXT8) on the first implementation of IESDR integration, I have added another argument Dt (analysis time-step) to Bilinear_Newmark_withTH function and closely followed the original MATLAB code.

Also removed the Cython implementation and reference to it for IESDR. It appears that the earlier performance optimisation effort has been made prematurely.

I'm copying Vahid's original feedback from Slack.

> @sung.bae (Cc @James) Firstly, thanks for implementing the function to the workflow. This part has been done perfectly. However, there are some small issues regarding the function.
> The most important one is that the part of the code checking the stability criteria of the numerical method is omitted in the python script. This part limits the time-step to ensure that the numerical solution remains stable (below lines). That is why the computed value for T=0.02 is very large in comparison to the others.
```
% Analysis time step
% ------------------
if isempty( Dt ) || Dt/min(T) > 0.551
   denominator = 2 * ceil( Dtg / min(Dtg/5,min(T)/30) / 2 );
   Dt = Dtg / denominator;
   disp( [ 'Note:  Dt set equal to ' num2str(Dt) ] );
end
```
> Another point that should be considered is that in the Matlab code two parameters were defined as time-step (Dt: analysis time-step and Dtg: time-step of waveform). This has been reduced to just one parameter in the Python version (dt, which is basically the time-step of waveform), which can reduce the precision of calculation when the time-step is large (e.g., dt=0.02 for this waveform). Usually, Dt is considered 0.005 for analysis regardless of the waveform time-step. Then, interpolation is used to calculate the other values. The Matlab script related to this (which is omitted in the Python version) is:
```
tg = [ 0:(length(ag)-1) ]' * Dtg;
t = [ 0:Dt:tg(end) ]';
p = interp1( tg, -ag*m, t );
```
> and the code calculates the inelastic displacement spectra, so maybe it is better to be named  “IESD”  instead of “IESDR”.


jpa198 (dismissed):

shot0829 (approved):

shot0829:
you might need to update the test, since you changed the vairable name to Dt, instead of dt

IM_calculation/IM/Burks_Baker_2013_elastic_inelastic.py
```diff
@@ -41,8 +41,19 @@
...
 
 def Bilinear_Newmark_withTH(
-    period: np.ndarray, z: float, dy: float, alpha: float, ag: np.ndarray, dt: float
+    period: np.ndarray,
+    z: float,
+    dy: float,
+    alpha: float,
+    ag: np.ndarray,
+    Dtg: float,
+    Dt: float,
```
jpa198, line 50: Should use lower case dt dtg, to be consistent with the rest of the codebase.
Consistency with ported source code is secondary (I think)

sungeunbae:
> you might need to update the test, since you changed the vairable name to Dt, instead of dt

Good point. Dt instead of dt doesn't matter, but there is an additional variable, so the test fails. Will fix tomorrow.
https://github.com/ucgmsim/IM_calculation/blob/master/IM_calculation/test/test_intensity_measures/test_elastic_inelastic.py

jpa198:
> > you might need to update the test, since you changed the vairable name to Dt, instead of dt
> 
> Good point. Dt instead of dt doesn't matter, but there is an additional variable, so the test fails. Will fix tomorrow.
> https://github.com/ucgmsim/IM_calculation/blob/master/IM_calculation/test/test_intensity_measures/test_elastic_inelastic.py

Either give it dtg, or allow dtg to have a default value (0.005 should be ok for most uses?)

IM_calculation/IM/Burks_Baker_2013_elastic_inelastic.py
shot0829: using dt, instead of Dt
pros: consistent with other steps we have
cons: if someone tried to search for variables after reading the doc for imported (and have no experience in our workflow) will have a hard time.

i would personally chose it to match the doc for ported source, because the most likely scenario for us to modify the code is when we have doubt and need to match it with original to find out the cause.

sungeunbae: Changed Dt -> dt 
There are bunch of other variables Ki, Ddi, Dpi etc keeping the same MATLABic style - but lets leave them out.

shot0829 (dismissed):

IM_calculation/IM/Burks_Baker_2013_elastic_inelastic.py
```diff
@@ -41,8 +41,19 @@
...
 ):
+    # Analysis time step
+    # If Dt is too high in relation to the period, adjust it to ensure numerical stability
+    if Dt is None or Dt / np.min(period) > 0.551:
```
jpa198: Dt should never be passed in as None. It should either have a default value of None or this test case should be removed

jpa198 (approved):

shot0829 (approved):

PR #97, opened 2021-03-16
Backarc

jasonmotha:


IM_calculation/scripts/calculate_rrups_single.py
```diff
@@ -66,43 +71,64 @@ def write_source_2_site_dists(
...
+    )
 
     args = parser.parse_args()
 
     # Load the stations
     station_df = load_station_file(args.station_file)
 
-    # Get the location for all station in the fd station file
+    # Matches the station file with the specified stations or the fd_stat_list. If neither is specified, use the whole list
     if args.fd_station_file:
```
claudio525, line 87: Could this whole block be simplified?
```python
station_mask = np.ones(station_df.shape[0], dtype=bool)
if args.fd_station_file:
    fd_stations_df = load_station_file(args.fd_station_file)
    station_mask = np.isin(station_df.index.values, fd_stations_df)
elif args.stations:
    station_mask = np.isin(station_df.index.values, args.stations)

filtered_station_np = station_df.loc[station_mask].values
stats_2_calc = station_df.loc[station_mask].index.values

# Or just use the dataframe directly?
station_df = station_df.loc[station_mask]
```
Something along those lines?

IM_calculation/source_site_dist/src_site_dist.py
```diff
@@ -132,3 +136,22 @@ def calc_rx_ry(srf_points: np.ndarray, plane_infos: List[Dict], locations: np.nd
             r_y[iloc] = np.min(np.abs([up_strike_dist, down_strike_dist]))
 
     return r_x, r_y
+
+
+def calc_backarc(srf_points: np.ndarray, locations: np.ndarray):
```
claudio525, line 141: docstring?

claudio525 (commented):
Whats backarc?

jasonmotha:
image

IM_calculation/scripts/calculate_rrups_single.py
jasonmotha: Sweet. The mask is a good idea. That makes it a wee bit simpler. 

I'd have preferred to use the df everywhere but I don't want to dig around and change too much as it's just a stylistic change rather than a readability change etc... Someone originally decided to use numpy for this script.

jpa198:
> image
> 
> @claudio This is backarc. Essentially classifying stations beyond the volcanic front.

In this case the backarc is mostly locations NW of Rotorua/Taupo given a subduction event on the east coast. Ie Hikurangi

jasonmotha:
Line definition (just uploading it here so I can link to this image in github)
untitled

jasonmotha:
Another note. This isn't multiprocessed, (neither is R_x / Ry0 either). So at some point on our todo list we should look at making both of these more performant. (numba style optimisation didn't work easily)

jpa198 (approved):

claudio525 (approved):

PR #98, opened 2021-03-23
Add h1 and h2 to the components list

jpa198:
Add function to return the currently available basic components
Refactor read_ascii_folder to load arbitrary basic components, including h1, h2
Formatting


jpa198:
I have verified that that the script runs and produces IMs for H1, H2. Any ideas on validation of the results?

jasonmotha:
Verification can be geom(000, 090) ~= geom(h1, h2) ?

jasonmotha (approved):
I refrain from making any comments as they fall back to the original author of these functions... Happy as long as validation makes sense.

jpa198:
> Verification can be geom(000, 090) ~= geom(h1, h2) ?

That didn't turn out as close as I'd hoped...

IM_calculation/IM/read_waveform.py
```diff
@@ -168,8 +168,9 @@ def read_ascii_folder(path, station_names, comp, units="g"):
                 c: open(os.path.join(path, f"{station}.{c.str_value}")) for c in comp
             }
         except IOError:
+            files_names = f"{station}.{{{[c.str_value for c in comp]}}}"
```
claudio525, line 171: Nitpicking here, but given that this variable is only used once. I suggest to either:
a) not define the variable and put the logic in the string formatting OR
b) define a variable, but put all the logic in its definition (i.e. currently the join is done in the string, and the rest is done in the variable definition), e.g. something like
```python
file_paths = os.path.join(path, f"{station}.{{{[c.str_value for c in comp]}}})
```

jpa198: Ah yeah, option b makes sense.
Option a includes more nesting " marks than I want to have to think about.

jpa198:
> > Verification can be geom(000, 090) ~= geom(h1, h2) ?
> 
> That didn't turn out as close as I'd hoped...

After discussion with Robin, and manual investigation, the H1, H2 outputs look alright. This PR is really only about loading them, and this seems to have worked

jasonmotha (approved):

shot0829 (approved):

claudio525 (approved):

PR #100, opened 2021-03-31
Station checkpoint

shot0829:
1. a few periods.out files that is required by adv_im models are also included in this PR. feel free to filter them out
2. a fix in im_calculation is also included


IM_calculation/Advanced_IM/runlibs_2d.py
```diff
@@ -41,6 +41,16 @@ def main(args, im_name, run_script):
...
+    for component in ["000", "090"]:
+        component_outdir = os.path.join(output_dir, component)
+        check_status(component_outdir)
+        station_completed = station_completed and check_status(component_outdir)
+    # skip the whole calculation if the station has been ran before
+    if station_completed:
+        print(f"{args} completed")
+        return
+
     model_converged = True
```
jpa198, line 54: If it doesn't converge on the first component, we don't need to run the second component

jpa198 (changes requested):

.gitignore
```diff
@@ -19,3 +19,6 @@ rrups.csv
 runs/
 **.so
 **sample0
+
+# special cases to unignore
+!periods.out
```
jasonmotha, line 24: Can't you autogenerate the periods.out file as part of setup rather than commiting huge lines of files?

shot0829: not sure, i'll ask Vahid how did he generate it.

shot0829: so the period.out should be treated as part of the model.
as the tcl loading, flags, and variable numbers are different between models.
I think renaming the extension to something else maybe the better solution here? ( so .gitignore doesnt has to change)

jasonmotha: Oh yeah. And the files are long but not actually that big so not a huge deal to store in git.

Would be good to know how they were generated though.

shot0829: i've asked vahid to walk me through.

turns out the periods.out is a product of running a modified version of "run_eigen_*models.tcl" under every adv_im_model/Original.
The tcl script provided currently under the folder for each model is merely a template. 
what he did was have a local copy of it and ran it once for each specific model, and then send the periods.out to me.

The reason doing so is that the period.out used to be part one full analysis, but for our case it will be generating the same data over and over again. 
So currently there are part of tcl scripts are never called because we skipped this 'pre-processing'

I spent sometime trying to run them, and most of them either has incorrect variable name or pathing.

I personally highly suggest we spend some time sorting them all out, and actually run this 'pre-processing' part as a install script/step. But i think that should be in another PR, since its going to be a huge effort going back and forth to make sure whatever local changes vahid had is reflected into the repository code base as well.

jasonmotha (approved):

jpa198 (approved):

PR #103, opened 2021-04-29
bugfix for estimating ascii file sizes

jasonmotha:


jpa198 (dismissed):
Off the top of my head an 8 digit number in ascii is about the same size as a double in numpy (np.float). So this works out

IM_calculation/IM/im_calculation.py
```diff
@@ -607,7 +607,13 @@ def get_steps(input_path, nps, total_stations, high_mem_usage=False):
...
     :return: number of stations per iteration/batch
     """
-    estimated_mem = os.stat(input_path).st_size * MEM_FACTOR
+    if os.path.isfile(input_path):
+        estimated_mem = os.path.getsize(input_path) * MEM_FACTOR
+    else:
+        names = os.listdir(input_path)
+        paths = [os.path.join(input_path, name) for name in names]
+        sizes = sum([os.path.getsize(path) for path in paths])
+        estimated_mem = sizes * MEM_FACTOR
```
claudio525, line 616: What about 
```python
sizes = sum([os.path.getsize(os.path.join(input_path, name)) for name in os.listdir(input_path)])
estimated_mem = sizes * MEM_FACTOR
```
or even just make it a one liner (i.e. remove `sizes` variable)?

claudio525 (approved):

claudio525 (approved):

jpa198 (approved):

PR #104, opened 2021-05-07
Move rotd func call to reduce ram use

jpa198:


jpa198:
Initial test on one waveform showed a 1/3rd reduction in peak ram usage.
Tests on full event ongoing

jasonmotha (approved):
Looks good. Makes sense too.

claudio525 (approved):

jpa198:
Ram use from the old way:
RotD50_old
Ram use with the new way:
RotD50_new
It capped out at points in both of them, but the new way does show less ram use
Test was run with RotD50 on a run of HikHBaymin

PR #105, opened 2021-05-12
Added EAS calculation as a separate component of FAS

sungeunbae:
EAS is a Euclidean distance of FAS 000 and 090 components. Instead of handing it as a separate IM of its own, we agreed to regard it as a special component of FAS.

If a user specifies to compute FAS (regardless of what components the user specifies), internally a component "EAS" is added to the comps_to_store list, and if the horizontal components 000 and 090 are not specified by the user, they will be added to the comps_to_calculate list. (Handled by the chance made to qcore, see the separate PR)

The code was tested under a few scenarios:
1. if user specifies FAS with 000 and 090 components, EAS is computed and the output csv contains 000, 090 and EAS
2. if user specifies FAS with 000 only, 090 is internally computed to produce EAS. The output csv contains 000 and EAS.
3. if user specifies FAS with no components, both 000 and 090 are internally computed to obtain EAS. The output csv contains only EAS (no geom)

Known issue: If user specifies FAS alongside other IMs, such as PGA, PGA will have an empty EAS component. FAS on the other hand will have an empty geom component.

### DIFF
--- a/IM_calculation/IM/im_calculation.py
+++ b/IM_calculation/IM/im_calculation.py
@@ -281,6 +281,11 @@ def calc_FAS(
             f"Attempting to compute fourier spectrum raised exception: {e}\nThis was most likely caused by attempting to compute for a waveform with more than 16384 timesteps."
         )
     else:
+        # compute EAS, the euclidean distance of FAS 000 and 090
+        values_to_store[
+            Components.ceas.str_value
+        ] = intensity_measures.get_euclidean_dist(value[:, 0], value[:, 1])
+
         for comp in comps_to_store:
             if comp.str_value in values_to_store:
                 for i, val in enumerate(im_options[im]):
@@ -437,7 +442,7 @@ def compute_measures_multiprocess(
 
     bbseries, station_names = get_bbseis(input_path, file_type, station_names)
     total_stations = len(station_names)
-    # determine the size of each iteration base on num of processes and mem
+    # determine the size of each iteration base on num of processers and mem
     steps = get_steps(
         input_path, process, total_stations, "FAS" in ims and bbseries.nt > 32768
     )
@@ -481,6 +486,10 @@ def compute_measures_multiprocess(
                     for waveform in waveforms
                 ]
                 all_results.extend(p.starmap(compute_measure_single, array_params))
+                # replace the line above with the below for debugging
+                # for array_param in array_params:
+                #     result=compute_measure_single(*array_param)
+                #     all_results.append(result)
 
     if running_adv_im:
         # read, agg and store csv
--- a/IM_calculation/IM/im_calculation.py
+++ b/IM_calculation/IM/im_calculation.py
@@ -59,7 +59,7 @@ def array_to_dict(value, comps_to_calc, im, comps_to_store):
             value_dict[comps_to_calc[i].str_value] = value[i]
     # In this case, if geom in str_comps,
     # it's guaranteed that 090 and 000 will be present in value_dict
-    if Components.cgeom in comps_to_store and im != 'FAS': #no geom for FAS
+    if Components.cgeom in comps_to_store:
         value_dict[Components.cgeom.str_value] = intensity_measures.get_geom(
             value_dict[Components.c090.str_value], value_dict[Components.c000.str_value]
         )
@@ -282,9 +282,10 @@ def calc_FAS(
         )
     else:
         # compute EAS, the euclidean distance of FAS 000 and 090
-        values_to_store[
-            Components.ceas.str_value
-        ] = intensity_measures.get_euclidean_dist(value[:, 0], value[:, 1])
+        if Components.ceas in comps_to_store:
+            values_to_store[
+                Components.ceas.str_value
+            ] = intensity_measures.get_euclidean_dist(value[:, 0], value[:, 1])
 
         for comp in comps_to_store:
             if comp.str_value in values_to_store:
--- a/IM_calculation/IM/im_calculation.py
+++ b/IM_calculation/IM/im_calculation.py
@@ -442,7 +442,7 @@ def compute_measures_multiprocess(
 
     bbseries, station_names = get_bbseis(input_path, file_type, station_names)
     total_stations = len(station_names)
-    # determine the size of each iteration base on num of processers and mem
+    # determine the size of each iteration base on num of processes and mem
     steps = get_steps(
         input_path, process, total_stations, "FAS" in ims and bbseries.nt > 32768
     )
@@ -486,10 +486,6 @@ def compute_measures_multiprocess(
                     for waveform in waveforms
                 ]
                 all_results.extend(p.starmap(compute_measure_single, array_params))
-                # replace the line above with the below for debugging
-                # for array_param in array_params:
-                #     result=compute_measure_single(*array_param)
-                #     all_results.append(result)
 
     if running_adv_im:
         # read, agg and store csv
--- a/IM_calculation/IM/im_calculation.py
+++ b/IM_calculation/IM/im_calculation.py
@@ -59,7 +59,7 @@ def array_to_dict(value, comps_to_calc, im, comps_to_store):
             value_dict[comps_to_calc[i].str_value] = value[i]
     # In this case, if geom in str_comps,
     # it's guaranteed that 090 and 000 will be present in value_dict
-    if Components.cgeom in comps_to_store:
+    if Components.cgeom in comps_to_store and im != 'FAS': #no geom for FAS
         value_dict[Components.cgeom.str_value] = intensity_measures.get_geom(
             value_dict[Components.c090.str_value], value_dict[Components.c000.str_value]
         )
--- a/IM_calculation/scripts/calculate_ims.py
+++ b/IM_calculation/scripts/calculate_ims.py
@@ -163,6 +163,21 @@ def load_args():
     )
 
     args = parser.parse_args()
+
+    if constants.Components.ceas.str_value in args.components:
+        wrong_ims = ""
+        if "FAS" not in args.im:
+            wrong_ims = ",".join(args.im)
+        else:
+            # there are IMs other than FAS but user only wants EAS
+            # without this check, the code still runs, but the non-FAS IMs
+            # won't be shown in the output csv without a notice/warning
+            if len(args.im) > 1 and len(args.components) == 1:
+                args.im.remove("FAS")
+                wrong_ims = ",".join(args.im)
+        if wrong_ims:
+            parser.error("EAS is not available with the specified IM types: {}".format(wrong_ims))
+
     calc.validate_input_path(parser, args.input_path, args.file_type)
     return args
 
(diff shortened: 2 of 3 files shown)


jasonmotha:
I would change it so that EAS is only returned if that component is specified? Otherwise, you don't have an option to turn it off if you don't want it.

sungeunbae:
> I would change it so that EAS is only returned if that component is specified? Otherwise, you don't have an option to turn it off if you don't want it.

I did ponder that option, and it's perhaps a more consistent behaviour in comparison to other IMs - except that user needs to know there is an unusual component EAS that needs to be specified.  I'm happy to change the behaviour, but I think we need to address this scenario. If user specifies FAS, but no component, we should add EAS like we add geom as a default to other IMs. (note that geom has no value when FAS is concerned)

jasonmotha:
The only reason geom is defaultly added is because it's specified as the default component - but we're starting to phase that towards rotd50. I think it's valid for a user to have to specify something additional if they want additional output.

We could potentially throw an argument error if EAS component is specified without FAS being present?

IM_calculation/IM/im_calculation.py
```diff
@@ -437,7 +442,7 @@ def compute_measures_multiprocess(
 
     bbseries, station_names = get_bbseis(input_path, file_type, station_names)
     total_stations = len(station_names)
-    # determine the size of each iteration base on num of processes and mem
+    # determine the size of each iteration base on num of processers and mem
```
jasonmotha, line 445: processors - but I think I prefer processes.

sungeunbae:
> The only reason geom is defaultly added is because it's specified as the default component - but we're starting to phase that towards rotd50. I think it's valid for a user to have to specify something additional if they want additional output.
>
I think the rotd50 blurrs the discussion point. Whatever the component is defined as the default, I think it is reasonable to keep the behavioural consistency.

- If a user specifies an IM but no component, geom is returned as the default. 
- If a user specifies FAS as an IM, but no component, EAS is returned as the default.
 
> We could potentially throw an argument error if EAS component is specified without FAS being present?
Good point. I didn't think of this case as the current behaviour doesn't need the user to specify EAS (it's automatically added).

My proposal (and compromise)
1. If user specified FAS, but desired components do not include EAS, don't include it in the output csv file.
2. If user specifies EAS as a component, but IM doesn't include FAS, throw an error.
3. If user specifies FAS as the only IM, but no component specified, use EAS as the default (like other IMs compute geom as the default.)
4. If user specifies FAS as the only IM, but only geom is specified, remove geom and use EAS as the default.
5. If user specified FAS alongside other IMs, but no component is specified, use EAS and geom as default for FAS and other IMs.
6. If user specified FAS alongside other IMs, and some components are specified, compute for those specified components.
7. If user specified FAS alongside other IMs, but only EAS is specified, use EAS and geom as default for FAS and other IMs.
8.If user specified FAS alongside other IMs, but only geom is specified, compute EAS as the default for FAS and geom for  other IMs. 

I think the current status already accommodates all above except 1,2.

IM_calculation/IM/im_calculation.py
```diff
@@ -481,6 +486,10 @@ def compute_measures_multiprocess(
                     for waveform in waveforms
                 ]
                 all_results.extend(p.starmap(compute_measure_single, array_params))
+                # replace the line above with the below for debugging
+                # for array_param in array_params:
+                #     result=compute_measure_single(*array_param)
+                #     all_results.append(result)
```
jasonmotha, line 492: Starmap debugs correctly in python3. It has both the starmap trace and the internal trace. Hence why we removed pool-wrapper. I don't think this is necessary to keep

IM_calculation/scripts/calculate_ims.py
```diff
@@ -123,9 +123,9 @@ def load_args():
         "--components",
         nargs="+",
         choices=list(constants.Components.iterate_str_values()),
-        default=[constants.Components.cgeom.str_value],
+        default= [constants.Components.cgeom.str_value],
         help="Please provide the velocity/acc component(s) you want to calculate eg.geom."
-        " Available compoents are: {} components. Default is all components".format(
+        " Available components are: {} components. Default is geom component".format(
```
jasonmotha, line 128: "Default is 'geom' or 'the geom component'"

jasonmotha (changes requested):
Looks good. Cool that there aren't many changes to get this implemented.

IM_calculation/IM/im_calculation.py
sungeunbae: Sure - pycharm went crazy, but I am happy to remove.

IM_calculation/IM/im_calculation.py
sungeunbae: Good catch.

jasonmotha:
The key part in the list is that IM_calc requires a component to be specified - you can't have "no" component.
The default is 'geom'. 
So 3,5,7 & 8 start to do some conditional logic which makes it more complicated. Considering the audience for our code, I think that having them explicitly specify what IMs they need + components is the best path forward. Otherwise we're introducing complexity into our code that we have to maintain - I think the simplicity is very useful.

sungeunbae:
> The key part in the list is that IM_calc requires a component to be specified - you can't have "no" component.
> The default is 'geom'.
> So 3,5,7 & 8 start to do some conditional logic which makes it more complicated. Considering the audience for our code, I think that having them explicitly specify what IMs they need + components is the best path forward. Otherwise we're introducing complexity into our code that we have to maintain - I think the simplicity is very useful.

3,4,5,6,7,8 are not additional conditional logic to maintain. They are possible use case scenarios and the current code already covers them with "not many changes". 

My point here is the default 'geom' is a wrong default for FAS.

However, I'm happy to produce a new version that just only does 1 and 2 above. it will not compute "EAS" unless specified.
If no component is specified, It will just return empty geom column (currently I suppressed it, as it is pointless) but I can turn it back on.

sungeunbae:
I made EAS be computed only when specified. If EAS is specified when FAS is not in the list, it throws an error. However, if user specifies FAS alongside other IM types, it gets tricky. If a user specifies EAS as the only desired component,  the default "geom" doesn't make to the component list.  As a result, the non-FAS IMs are silently removed from the output csv. (as it has no component to compute).
This version just throws an error if it is the case.
Alternatively, we can add "geom" to the list, and carry on. 
I think you will favour the former, which is simpler.

IM_calculation/scripts/calculate_ims.py
```diff
@@ -163,6 +163,21 @@ def load_args():
...
+            wrong_ims = ",".join(args.im)
+        else:
+            # there are IMs other than FAS but user only wants EAS
+            # without this check, the code still runs, but the non-FAS IMs
+            # won't be shown in the output csv without a notice/warning
+            if len(args.im) > 1 and len(args.components) == 1:
+                args.im.remove("FAS")
+                wrong_ims = ",".join(args.im)
+        if wrong_ims:
+            parser.error("EAS is not available with the specified IM types: {}".format(wrong_ims))
```
jasonmotha: I think it might be useful to Include how to rectify. "adding FAS as a specified IM will allow EAS to be calculated." or...
But it might make the error message quite long.

IM_calculation/IM/im_calculation.py
```diff
@@ -282,9 +282,10 @@ def calc_FAS(
         )
     else:
         # compute EAS, the euclidean distance of FAS 000 and 090
-        values_to_store[
-            Components.ceas.str_value
-        ] = intensity_measures.get_euclidean_dist(value[:, 0], value[:, 1])
+        if Components.ceas in comps_to_store:
+            values_to_store[
+                Components.ceas.str_value
+            ] = intensity_measures.get_euclidean_dist(value[:, 0], value[:, 1])
```
jasonmotha, line 288: Does this only calculate EAS for FAS? It looks like it calculated the euclidian distance for all IMs? But I'm happy to be proven wrong.

jasonmotha (changes requested):
Some minor comments. Looks good.

IM_calculation/IM/im_calculation.py
sungeunbae: Not sure of your question. This is a part of calc_FAS function, and computes Euclidean distance for FAS.

IM_calculation/scripts/calculate_ims.py
```diff
@@ -176,7 +176,7 @@ def load_args():
                 args.im.remove("FAS")
                 wrong_ims = ",".join(args.im)
         if wrong_ims:
-            parser.error("EAS is not available with the specified IM types: {}".format(wrong_ims))
+            parser.error("The specified IMs need non-EAS components to proceed: {}".format(wrong_ims))
```
sungeunbae, line 179: @jasonmotha You didn't check the latest error message

IM_calculation/scripts/calculate_ims.py
sungeunbae: I have already revised the error message, which implies how to resolve. See below.

jasonmotha: Perfect. That encomapasses both cases.

That'll teach me for reading commits individually.

IM_calculation/IM/im_calculation.py
jasonmotha: Oh it does too. I see how that works now. I presumed it was in the part near where geom was calculated.

IM_calculation/scripts/calculate_ims.py
sungeunbae: Sorry about the confusion, I changed my mind after making one commit.

shot0829 (approved):

IM_calculation/scripts/calculate_ims.py
```diff
@@ -163,6 +163,25 @@ def load_args():
     )
 
     args = parser.parse_args()
+
+    if constants.Components.ceas.str_value in args.components:
+        wrong_ims = ""
+        if "FAS" not in args.im:
```
jpa198, line 169: Using enum str value above, but a string here is inconsisent. 
I think this should use the enum str value

IM_calculation/scripts/calculate_ims.py
```diff
@@ -163,6 +163,25 @@ def load_args():
...
+        else:
+            # there are IMs other than FAS but user only wants EAS
+            # without this check, the code still runs, but the non-FAS IMs
+            # won't be shown in the output csv without a notice/warning
+            if len(args.im) > 1 and len(args.components) == 1:
+                args.im.remove("FAS")
+                wrong_ims = ",".join(args.im)
+        if wrong_ims:
+            parser.error(
+                "The specified IMs need non-EAS components to proceed: {}".format(
```
jpa198, line 180: We should say that EAS needs FAS if we're going to go down this route.
However I think this should be handled more like components.
If FAS is given we calculate and save FAS.
If ESA is given we calculate FAS, use it to compute EAS and save only that.
If both are given we calculate and save both.

jpa198 (changes requested):

jpa198:
Uh, I didn't see the conversation that happened before I reviewed this. I thought the discussion we had on slack was determining how this would be implemented. Will have a read here and go over slack again.

IM_calculation/scripts/calculate_ims.py
sungeunbae: EAS is a separate component, only used by FAS.
>If FAS is given we calculate and save FAS: Yes for the specified component. 
>If ESA is given we calculate FAS, use it to compute EAS and save only that: No, EAS is not a stand-alone. If EAS is specified as a desired component but FAS is not in the IM list, an error is thrown.
>If both are given we calculate and save both: Yes, It will compute EAS component of FAS IM type.

jpa198 (approved):
After going over it again, I think this is ok

jasonmotha (approved):

PR #107, opened 2021-05-24
If cython version is not available, import python version of rspectra

sungeunbae:
This has never been scoped, but I thought this could come in handy if a user doesn't have a working environment to compile Cython version of rspectra.


jasonmotha:
I'm personally not that keen to keep two copies of rspectra functions. Just a larger potential for bugs as we need to edit both files. I don't think we've had cython issues since we combined the setup.py scripts.

sungeunbae:
> I'm personally not that keen to keep two copies of rspectra functions. Just a larger potential for bugs as we need to edit both files. I don't think we've had cython issues since we combined the setup.py scripts.

Vahid with Windows PC can't compile cython (no permission to get a Linux subsystem) and has no means to install IM_calculation.  Travis and now Jenkins environment (with minimal linux installation) had to carry pre-compiled cython lib around, which is pretty clunky.

I agree having two copies may feel uncomfortable. But rspectra code is very stable, we rarely touch it.
Alternatively we can repalce cython with numba - but that's a lot more significant change. Can't we just have this in, and converge into a better alternative later?

sungeunbae:
@jasonmotha BTW ,my preliminary investigation, 

(base) (venv) sungbae@~/IM_calculation$ PYTHONPATH=~/qcore:~/IM_calculation time python IM_calculation/scripts/calculate_ims.py -i cython -m pSA -p 1.0 0.5 0.25 -t s ~/FAS-Calculation/Hossack/Hossack_REL10/BB/Acc/BB.bin b -o /Users/sungbae/tmp -c 090 000 ver
Reading waveforms in: g
Calculations are outputted to /Users/sungbae/tmp

Cython version:       20.27 real         5.16 user         1.61 sys
Numba version:      21.50 real        10.20 user         1.93 sys
Python version:      99.26 real        98.45 user         2.15 sys

and the output are exactly the same.

jpa198:
Can this have numba added to it in this PR? 
There are still concerns with duplicate code. But if this with numba has the same performance as the cython, we might be able to refactor out the cython eventually?

sungeunbae:
> Can this have numba added to it in this PR?
> There are still concerns with duplicate code. But if this with numba has the same performance as the cython, we might be able to refactor out the cython eventually?

Sure - I'll be testing Numba's compatibility (it relies on LLVM) and see if it works on NeSI and KISTI.

sungeunbae:
Numba-jit wants to use its own version of np functions, and data types for best result.
Previously I just used `@numba.jit` but it is recommended to use `@numba.jit(nopython=True)` or `@numba.njit`, which wouldn't work as `np.zeroes(..dtype=float)` runs into an error. The fix is to use numba's own float, namely, numba.float64 instead of float. This easily doubles the speed.

jasonmotha:
Sweet - numba version generally looks good. I'm happy with the numba approach replacing the cython. Though, we will need a wiki page to show that the answers between the two approches are the same.
And a test for a medium and large fault for performance too (say Moonshine / AlpineF2K), ideally 10 runs and averaged.

As a note to your comment of merge this in and change it later. Unless there is a specific plan to change it later, I do not believe it would get done. 
The other thing is, this change isn't scoped as part of an epic / sprint at the moment. So, might be a better idea to shelve this entirely until next sprint? Or if we deem it to be high-priority we can push it through.

WRT to vahid, rather than us having to support windows. He should ask Olive to either convert his computer to linux  (or give him permission to do so) or ask to install linux subsystem (but then you end up with issues like Mike) or (IMO the best solution) install linux in a VM for him - I'm pretty sure you can get a VM manager from the Software Centre.

sungeunbae:
> Sweet - numba version generally looks good. I'm happy with the numba approach replacing the cython. Though, we will need a wiki page to show that the answers between the two approches are the same.
> And a test for a medium and large fault for performance too (say Moonshine / AlpineF2K), ideally 10 runs and averaged.
> 
> As a note to your comment of merge this in and change it later. Unless there is a specific plan to change it later, I do not believe it would get done.
> The other thing is, this change isn't scoped as part of an epic / sprint at the moment. So, might be a better idea to shelve this entirely until next sprint? Or if we deem it to be high-priority we can push it through.

That's why I was suggesting to push the plain python version this time. Quick and easy. It will create a momentum to revisit this issue next Sprint, and I can do some thorough benchmark/validation.
 
> WRT to vahid, rather than us having to support windows. He should ask Olive to either convert his computer to linux (or give him permission to do so) or ask to install linux subsystem (but then you end up with issues like Mike) or (IMO the best solution) install linux in a VM for him - I'm pretty sure you can get a VM manager from the Software Centre.

Maybe. It feels like getting a bazooka when all I need is a flyswatter, though. I think Numba approach will be a good compromise.

Well, I am stopping pushing this. Let's sit on this issue and revisit next Sprint and plan+rescope (Validation epic?)

My suggestion here: 
1. If Numba version works perfectly, Cython will go. We only keep Numba.
3. If Numba version doesn't work on HPC, let's keep both Cython and Numba.
4. If we decide Numba is not good, keep Cython and plain Python as a fall-back option.

sungeunbae:
Just personal note: Numba works on Maui - had a number of issues initially, but the combo of  Numba 0.47 + llvmlite 0.31. worked ok. It is not as fast as Cython version. (36sec vs 14sec), but the results are identical (not even floating point calc difference)

jpa198:
Waiting on formatting change and an agreement on how to move forward with cython vs numba'd python. 
I'm not a big fan of numba (It never seems to work for me), but I don't really want two copies of this in the repo either.
Could we just a keep a copy of the python version somewhere and only use it for Vahid on windows? Otherwise I'm inclined to go for numba if it works for everyone else

sungeunbae:
> Waiting on formatting change and an agreement on how to move forward with cython vs numba'd python.
> I'm not a big fan of numba (It never seems to work for me), but I don't really want two copies of this in the repo either.
> Could we just a keep a copy of the python version somewhere and only use it for Vahid on windows? Otherwise I'm inclined to go for numba if it works for everyone else

If I have to choose between Cython and Numba, I'm inclined to go with Numba - its low overhead is something Cython can't match, and it has come a long way to support an ordinary Python interpreter (it uses LLVM to JIT compile, but doesn't need the interpreter to be LLVM-built, which helps its HPC availability). Having said that, the performance on Maui is a bit shoddy based on my preliminary investigation. My preferred option is getting Numba working, and keep it as the only copy. Otherwise, as James indicated, we can just go back to Cython with a Python copy kept as a fall-back option.

jpa198:
Which epic is this being assigned to? Platforms? Seeing as platform compatibility is why it came up in the first place

jpa198:
Do we want to do anything with this? I don't think we really need it with the variety of linux machines available to run on

sungeunbae:
Sure - With Vahid having submitted his thesis, we can live without this.

PR #112, opened 2021-06-17
Update verify adv im

shot0829:


IM_calculation/Advanced_IM/runlibs_2d.py
```diff
@@ -12,6 +14,13 @@
 DEFAULT_OPEN_SEES_PATH = "OpenSees"
 DF_INDEX_NAME = "component"
 
+TIME_FORMAT = "%Y-%m-%d %H:%M:%S"
+
+
+class time_type(Enum):
+    start_time = 0
+    end_time = 1
```
jpa198, line 22: If these aren't being imported then just having them as globals is fine.
Given how you're using them (for the name only) having an entire enum seems like overkill.

shot0829: the enum is used in the verify_adv_im.py in gmsim_workflow.

jpa198: Ok, they are being imported by the other PR. This is fine

jpa198 (dismissed):

IM_calculation/Advanced_IM/runlibs_2d.py
jasonmotha: We're trying to limit importing from IM_calc(and Pre-processing among others) to other repos - for circular dependancies. Perhaps this enum is better off in qcore / slurm workflow? (But haven't read much of the PR yet)

IM_calculation/Advanced_IM/runlibs_2d.py
```diff
@@ -44,6 +53,15 @@ def parse_args():
     return args
 
 
+def datetime_to_file(t_value, t_type, out_dir):
```
jasonmotha, line 56: Should be clear that this is the string not the enum.

jasonmotha (changes requested):

IM_calculation/Advanced_IM/runlibs_2d.py
shot0829: this verify script is already importing functions from runlib2d.py . moving this to other places may increases the dependancies, instead of reducing it.

jasonmotha: Thoughts on moving the verify script into IM_calc rather than having it part of workflow? It doesn't seem to have much slurm dependancies? @jpa198

shot0829: my original reason for putting the verify script in workflow is because all other verification codes are in there as well (emod3d, hf, bb). and since Adv_IM is part of the automated workflow, i thought this makes it more consistent.
However, moving it from workflow to IM_calc also removes one dependency/connection.

jasonmotha: That makes sense for hf/bb because their scripts (arguably incorrectly) are there too. 
EMOD is a different question entirely - but all the support scripts for emod are in workflow too.

jpa198: Yeah, that should be in IM_calc rather than slurm_workflow

jasonmotha:
https://wiki.canterbury.ac.nz/display/QuakeCore/Advanced+IM+status+script

Tying documentation to the PR. Documentation looks good.

jpa198 (approved):
Yeah, looks good I think

IM_calculation/Advanced_IM/check_adv_IM_status.py
```diff
@@ -0,0 +1,230 @@
...
+    parser.add_argument(
+        "adv_im_model", nargs="+", type=str, help="list of adv_IM models that ran"
+    )
+
+    parser.add_argument(
+        "--components",
+        default=[const.Components.c000.str_value, const.Components.c090.str_value],
+        nargs="+",
+        choices=list(const.Components.iterate_str_values()),
+        help="list of component that ran",
```
jasonmotha, line 66: components

jasonmotha (approved):

IM_calculation/Advanced_IM/check_adv_IM_status.py
shot0829: can i merge this PR and put this fix in the next one XD

jasonmotha: Yep. It's approved. =)

PR #115, opened 2021-07-14
Adv im fix

shot0829:


IM_calculation/scripts/calculate_ims.py
```diff
@@ -206,7 +206,7 @@ def main():
 
     # Create output dir
     utils.setup_dir(args.output_path)
-    if not args.simple_output:
+    if not args.simple_output and args.advanced_ims is None:
```
shot0829, line 209: just a note, this was added because the verification on observed workflow is using glob to determine the station list.
so creating a 'stations' folder with nothing in side causes the check_status to always return failed.

IM_calculation/Advanced_IM/runlibs_2d.py
```diff
@@ -111,7 +111,10 @@ def main(args, im_name, run_script):
         )
 
         try:
-            subprocess.run(script, timeout=int(args.timeout_threshold))
+            subprocess.run(
+                script,
+                timeout=int(args.timeout_threshold) if args.timeout_threshold else None,
```
jpa198, line 116: Surely this if statement could be avoided by setting the argument type to int and default to None?

sungeunbae: if default is None, int(None) will run into TypeError

shot0829: you are correct. updated.

IM_calculation/Advanced_IM/runlibs_2d.py
```diff
@@ -89,7 +92,7 @@ def main(args, im_name, run_script):
                 f"{getattr(args,component.str_value)} failed to converged in previous run."
             )
             model_converged = False
-            break
+            continue
```
sungeunbae, line 95: I think you have to update the comment and clarify what the code is trying to do here

IM_calculation/Advanced_IM/runlibs_2d.py
```diff
@@ -127,7 +130,6 @@ def main(args, im_name, run_script):
                 f"{component_outdir} failed to converge, skipping rest of the components"
             )
             model_converged = False
```
sungeunbae, line 132: Comment here and describe you are still computing even if one component fails - It'll be worth stating that potentially geom component can't be worked out.

IM_calculation/scripts/calculate_ims.py
sungeunbae, line 209: Is the default value for nargs="+" None? Just checking.

IM_calculation/Advanced_IM/runlibs_2d.py
shot0829: about the int part, i've updated the code base on james comment.

sungeunbae: So that. That will work.

IM_calculation/scripts/calculate_ims.py
shot0829: yes, it will be None not a empty list or such.

jasonmotha: I think that this isn't a robust solution to your problem. Expecting nothing else to be in the folder isn't a great solution. Could you get the observed station list from the input waveforms? or potentially write a list of stations as part of IM_calc?

jasonmotha (changes requested):

IM_calculation/scripts/calculate_ims.py
shot0829: the current way of getting waveforms is from scanning the entire folder for *.000, instead of reading a file that contains a list. 
so we can potenailly generate a temporary file that contains that list and use that as the input to check_status?

jasonmotha: Yep. That sounds good. A permanent file rather than a temporary one but sure.

sungeunbae (approved):

IM_calculation/scripts/calculate_ims.py
shot0829: https://github.com/ucgmsim/slurm_gm_workflow/pull/386/commits/654a5280f397e63922fecb6713abf7e38cdd8b9b
i've update the script to just use the get_bbseis function from im_calc.
so on files needs to be saved, as its using the same logic

IM_calculation/Advanced_IM/runlibs_2d.py
```diff
@@ -138,11 +142,12 @@ def main(args, im_name, run_script):
         # check for success message after a run
         # marked as failed if any component fail
         if not check_status(component_outdir):
+            # even if not conveged, script should still run other compoents, except geom.
```
jasonmotha, line 145: converged* components*

jasonmotha (approved):
Some spelling issues but otherwise happy.

shot0829:
doing a quick test run. should be one more patch coming in

sungeunbae (dismissed):
I'm happy - Just need to make @jasonmotha happy.

IM_calculation/Advanced_IM/runlibs_2d.py
sungeunbae: keep @jasonmotha happy

shot0829:
Test done. completed ATC_3story without error using opensees_advanced_im.sh.

sungeunbae (dismissed):

PR #116, opened 2021-07-16
Created SDI as an Advanced IM

sungeunbae:
- SDI is still available as a basic IM
- This PR adds SDI to Advanced IMs: Will give an option to researchers to freely play with parameters. Largely followed how T&M 2005 was implemented.


jpa198 (approved):

IM_calculation/IM/Burks_Baker_2013_elastic_inelastic.py
```diff
@@ -132,7 +134,9 @@ def Bilinear_Newmark_withTH(
     # Ha = a
     # Hfs = fs
 
-    return d  # return all displacements, not just the maximum Sd. This is to compute rotd if needed
+    return (
+        d
+    )  # return all displacements, not just the maximum Sd. This is to compute rotd if needed
```
jpa198, line 139: This seems to be failing the black test

sungeunbae: It seems to think my black is not black enough - fixed.

jpa198 (approved):

IM_calculation/Advanced_IM/runlibs_2d.py
```diff
@@ -27,15 +27,15 @@ def parse_args():
     parser = argparse.ArgumentParser()
 
     parser.add_argument(
-        Components.c000.str_value,
+        f"comp_{Components.c000.str_value}",
```
shot0829, line 30: is there a reason behind this change? because doing so will likely break codes like 
https://github.com/ucgmsim/IM_calculation/blob/9092fc3ee869d220a053ad3882e95456cfbe4928/IM_calculation/Advanced_IM/runlibs_2d.py#L102

shot0829 (changes requested):

IM_calculation/IM/intensity_measures.py
```diff
@@ -58,12 +58,11 @@ def get_SDI_nd(acceleration, period, NT, DT, z, alpha, dy, dt):
     if acceleration.ndim != 1:
         ts, dims = acceleration.shape
         Nstep = calculate_Nstep(DT, NT)
-        displacements = np.zeros((period.size, Nstep - 1, dims))
+        displacements = np.zeros((period.size, Nstep, dims))
```
shot0829, line 61: what was the reason for changing the shape? 
i have a vague memory that it was set to Nt -1 for a reason

IM_calculation/Advanced_IM/runlibs_2d.py
sungeunbae: run.py (not just SDI, see run.py@Taghavi_Miranda_2005) calls runlibs_2d.parse_args() and expects to get their arguments to be like args.comp_000, args.comp_090.

IM_calculation/IM/Burks_Baker_2013_elastic_inelastic.py
```diff
@@ -68,7 +68,9 @@ def Bilinear_Newmark_withTH(
     # % Interpolate p=-ag*m (linearly)
     p = -ag * m
     tg = np.linspace(0, ag.size - 1, ag.size) * dtg
-    t = np.linspace(0, tg[-1], int(tg[-1] / dt) + 1)  # num of steps increased from tg
+    t = np.linspace(
+        0, tg[-1], int(np.ceil(tg[-1] / dt)) + 1
```
sungeunbae, line 72: The division here sometimes caused poor rounding issue and will cause t and tg to have occasional size difference of 1. np.ceil() prevents the issue.

IM_calculation/IM/intensity_measures.py
sungeunbae: Thought Nstep -1 should be correct as it is "displacements". It turned out sometimes it can be Nstep, and other times Nstep - 1. 
It was due to rounding. I added comment in Burks_Baker_2013_elastic_inelastic.py

IM_calculation/Advanced_IM/runlibs_2d.py
sungeunbae: If this is only specific to SDI and T&M 2005, I'm happy to revert the change to runlibs_2d and fix SDI and T&M run.py instead.

sungeunbae: @jasonmotha Your thoughts?

shot0829: the reason it was coded this way is to avoid local definition of Components, and use the ones from qcore.
so instead of a local
https://github.com/ucgmsim/IM_calculation/blob/5de0f138b4e971ca902dcf96beafc3a631ec7264/IM_calculation/Advanced_IM/Models/Burks_Baker_2013/run.py#L22
its using 
https://github.com/ucgmsim/IM_calculation/blob/9092fc3ee869d220a053ad3882e95456cfbe4928/IM_calculation/Advanced_IM/runlibs_2d.py#L78
and to extract the value from args using 
https://github.com/ucgmsim/IM_calculation/blob/9092fc3ee869d220a053ad3882e95456cfbe4928/IM_calculation/Advanced_IM/runlibs_2d.py#L102

shot0829: i would prefer a centralized string and enum. thats why it was coded that way

IM_calculation/IM/intensity_measures.py
shot0829: have you tested this on ASCII data?
because last time similar thing happened and it was because of the float32 and float64 precision difference.
the solution was to kept the old shape but force the dt into float32
see :
https://github.com/ucgmsim/IM_calculation/blob/ffd4997c399296b91c8ef1afe0b5cb8d2107eb20/IM_calculation/IM/read_waveform.py

jasonmotha (changes requested):

IM_calculation/IM/intensity_measures.py
jasonmotha: Testing with ASCII is a good call, Jesse encountered some bugs due to Nstep - 1 and that's the reason why it is what it is in pSA at the moment.

sungeunbae: Yes, tested with both binary and ASCII.

IM_calculation/Advanced_IM/Models/Burks_Baker_2013/run.py
```diff
@@ -0,0 +1,94 @@
...
+        for dt in dt_list:
+            for alpha in alpha_list:
+                for dy in dy_list:
+                    displacements = intensity_measures.get_SDI_nd(
+                        accelerations, period, NT, DT, z, alpha, dy, dt
+                    )
+                    sdi_values = np.max(np.abs(displacements), axis=1)
+                    im_names = []
+                    for t in period:
+                        im_name = f"SDI_{t}_dy{dy}"  # we only vary dy at the moment
```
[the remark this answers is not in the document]
sungeunbae: Vahid specifically requested this to be done this way

IM_calculation/IM/Burks_Baker_2013_elastic_inelastic.py
sungeunbae: Already done in the latest commit

IM_calculation/Advanced_IM/runlibs_2d.py
sungeunbae: OK - Fixed as per @shot0829 's suggestion. Also fixed T&M 2005.

IM_calculation/Advanced_IM/Models/Burks_Baker_2013/run.py
```diff
@@ -0,0 +1,94 @@
+"""
+Python script to run a 3D waveform through Taghavi_Miranda_2005 and store the outputs to a txt file
```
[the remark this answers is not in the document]
sungeunbae: LOL

IM_calculation/Advanced_IM/Models/Burks_Baker_2013/run.py
```diff
@@ -0,0 +1,94 @@
...
+                        if t not in ordered_columns_dict:
+                            ordered_columns_dict[t] = []
+                        ordered_columns_dict[t].append(im_name)
+
+                    for i, component in enumerate(COMPONENT):
+                        if component not in results:
+                            results[component] = {}
+                        for j, im_name in enumerate(im_names):
+                            results[component][im_name] = (
+                                sdi_values[j, i] * 100
```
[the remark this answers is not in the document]
sungeunbae: Done

IM_calculation/Advanced_IM/Models/Burks_Baker_2013/run.py
```diff
@@ -0,0 +1,94 @@
...
+        ordered_columns.extend(
+            ordered_columns_dict[t]
+        )  # order by period first, then by dy
+
+    im_csv_fname = os.path.join(output_dir, "Burks_Baker_2013.csv")
+    df = pd.DataFrame.from_dict(results, orient="index")
+    df.index.name = "component"
+    geom = pd.Series(get_geom(df.loc["000"], df.loc["090"]), name="geom")
+    df = df.append(geom)
+    df[ordered_columns].to_csv(im_csv_fname)
```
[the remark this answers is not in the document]
sungeunbae: I am really not sure what this means..?

shot0829 (dismissed):

jpa198 (approved):
Once Jasons comment is resolved I think this looks good to go

jpa198 (dismissed):

IM_calculation/IM/intensity_measures.py
```diff
@@ -58,10 +58,14 @@ def get_SDI_nd(acceleration, period, NT, DT, z, alpha, dy, dt):
     if acceleration.ndim != 1:
         ts, dims = acceleration.shape
         Nstep = calculate_Nstep(DT, NT)
```
jpa198, line 60: Is Nstep needed anymore?

sungeunbae: Good point - please review my latest change

IM_calculation/IM/Burks_Baker_2013_elastic_inelastic.py
```diff
@@ -67,8 +67,13 @@ def Bilinear_Newmark_withTH(
...
     p = -ag * m
-    tg = np.linspace(0, ag.size - 1, ag.size) * dtg
-    t = np.linspace(0, tg[-1], int(tg[-1] / dt) + 1)  # num of steps increased from tg
+
+    tg = np.arange(
+        0, ag.size * dtg, dtg, dtype=float
+    )  # Creating [ 0, dtg, 2*dtg.... (ag.size-1)*dtg ].
+    t = np.arange(
+        0, ag.size * dtg, dt, dtype=float
+    )  # For the same begin and end, create a range spaced by dt. num of steps increased from tg
```
sungeunbae, line 76: I think this is a lot more intuitive than the original code. This is also more reliable we don't do the division to work out the number of steps (unlike linspace())

jpa198: Reading the documentation, using non integer step size with arange is often not consistent.
It may be better to calculate the number of timesteps and then multiply by the timestep size.

https://numpy.org/doc/stable/reference/generated/numpy.arange.html

jpa198 (approved):

IM_calculation/IM/Burks_Baker_2013_elastic_inelastic.py
jpa198: We may need to keep an eye on this going forward, but it should be fine

sungeunbae: To see why the original code is not correct, try:
ag.size = 7500, dtg=0.02, dt=0.005
tg.size is 7500, and t.size is 29996, while it should be 30000.
The second linspace could have been
`tg=np.linspace(0,(ag.size-1)*dtg,ag.size)
t=np.linspace(0,ag.size*dtg-dt,int(np.ceil((ag.size*dtg)/dt)))`
but as you can see, linspace() is not very readable - compare with the equivalent arange()

sungeunbae: It's because linkspace()'s second argument is the last element (inclusive). As we wanted ag.size*dtg to be the upperbound, we put (ag.size-1)*dtg for tg. For t, we also want ag.size*dtg to be the upperbound, but as the second argument is the last element, it should be ag.size*dtg-dt. If we want to use tg[-1] as in the original code, it should have been 
`
t=np.linspace(0,tg[-1]+dtg-dt, int(np.ceil( (tg[-1]+dtg)/dt)))
`

sungeunbae: @jpa198  You win. Came across a case where arange() fails and goes over the specified value of "stop". Reverted to linspace(), but left the arange() equivalents as comment to make it easier to understand

jpa198 (dismissed):

IM_calculation/Advanced_IM/Models/Burks_Baker_2013/run.py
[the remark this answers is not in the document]
sungeunbae: @jasonmotha  Welcome back and could you have a look at this PR again?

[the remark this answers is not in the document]
jpa198: @jasonmotha ?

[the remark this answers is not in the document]
sungeunbae: @jasonmotha ?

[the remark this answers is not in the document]
jpa198: Because each value in ordered_columns is f"SDI_{t}_dy{dy}", it doesn't allow for multiple dt, z or alpha values, which are currently taken from a list. We need to go back to Vahid to find out how he wants to handle multiple dt, z and alpha values. 
If he doesn't, then we either need to make dt, z and alpha single value, or put them in the im name anyway and make him a post processing script to strip them out.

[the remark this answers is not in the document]
sungeunbae: Good point - he is still not sure if he will need multiple values. The current structure gives him an option to try things out. Valid point on the column names.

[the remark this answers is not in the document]
sungeunbae: @jpa198  Added a change to show all variables in column names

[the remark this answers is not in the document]
jpa198: @jasonmotha is it acceptable now?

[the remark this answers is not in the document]
jasonmotha: @jpa198 / @sungeunbae Any reason for the explicit ordering (storing the order in a list) rather than just sorting the column names afterwards?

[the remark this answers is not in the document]
sungeunbae: @jasonmotha I don't remember the details - it was written quite a while ago, but I showed @jpa198 when I had a great deal of trouble with the sorting function to keep the sensible numeric/ alphabetical order.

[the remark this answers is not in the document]
jasonmotha: I think you showed me that too; wasn't the end of that issue that the df was getting written twice? Once was sorted, and then the second time wasn't and kept overwriting the sorted one?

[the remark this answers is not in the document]
sungeunbae: No not that one. That one was a case where no change of order seen before and after. This one was where the order was mixed up with the available sorting method when alphabet and numeric values are all there. You know 1, 10, 2..when you want 1,2,10. I tried natural sort and other ways and nothing worked. Enlighten me.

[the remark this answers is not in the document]
shot0829: I haven't seen the exact csv so not sure if this solution works for you as well:
https://github.com/ucgmsim/IM_calculation/blob/36f25f00157fb3382570c36a6682a2002a3d6bc8/IM_calculation/Advanced_IM/advanced_IM_factory.py#L156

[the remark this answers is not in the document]
sungeunbae: I tried that too and didnt work. cant remember the details. It's been 20 days and I never thought this was going to be a hot talking point. One thing I remember, a simple sort worked better for these column names than this natural sort, and it almost worked, except this example. Uploading csv shortly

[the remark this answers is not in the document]
sungeunbae: [Burks_Baker_2013.csv](https://github.com/ucgmsim/IM_calculation/files/6919907/Burks_Baker_2013.csv)
Screen Shot 2021-08-03 at 9 24 03 AM

IM_calculation/Advanced_IM/Models/Burks_Baker_2013/run.py
```diff
@@ -0,0 +1,140 @@
...
+                    )
+                    sdi_values = np.max(np.abs(displacements), axis=1)
+                    im_names = []
+                    for t in period:
+                        im_name = f"SDI_{t}_dy{dy}_a{alpha}_dt{dt}_z{z}"
+                        im_names.append(im_name)
+                        # for each period, we will have a list of im_names varying dy,alpha,dt and z, to help ordering columns later
+                        if t not in ordered_columns_dict:
+                            ordered_columns_dict[t] = []
+                        ordered_columns_dict[t].append(im_name)
```
jasonmotha, line 80: my favourite way of doing this block is ```ordered_columns_dict.get(t, []).append(im_name).```

But happy as is; it seems that there are a few instances where you do the check, create, update paradigm.

IM_calculation/Advanced_IM/Models/Burks_Baker_2013/run.py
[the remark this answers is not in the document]
jasonmotha: My main concern was to reduce complexitity by leveraging built-ins rather than having to keep track of it manually, if it's not working easily. Then happy as is.

jasonmotha (approved):

jpa198 (dismissed):

jpa198:
Yeah, I've just made another side branch to muck around in to fix the qcore issue...

jpa198 (approved):

jasonmotha (approved):

shot0829 (approved):

PR #117, opened 2021-07-19
Added real stations only option

sungeunbae:


shot0829 (approved):

IM_calculation/scripts/calculate_ims.py
```diff
@@ -136,6 +136,13 @@ def load_args():
         type=int,
         help="Please provide the number of processors. Default is 1",
     )
+
+    parser.add_argument(
+        "--real_stats_only",
+        action="store_true",
+        help="Please add '-r' to indicate if you want to consider real stations only",
```
jpa198, line 143: An explicit '-r' would be nice, but should be covered by argparse automatically adding abbreviations.

jpa198 (approved):

IM_calculation/scripts/calculate_ims.py
sungeunbae: Conflicts with -r --rupture

jpa198: Then the comment is wrong as adding -r is for --rupture, and will cause a crash if someone does try -r for --real_stats_only. Can you change the comment and matching documentation?

sungeunbae: Thanks for catching this!

jasonmotha:
Just for my curiosity, what was the use case for this feature?

shot0829:
> Just for my curiosity, what was the use case for this feature?

Some simulation runs on validation events are ran using non-uniform-grid, instead of just real stations.
The automated workflow or script i created will map stations from observed and bypass this issue, but if you want to manual submit job it will be troublesome .

shot0829:
this change also removes the dependency on having a observed run. so even if you want to just run analysis on a simulated data that has virtual-stations, this change will make it a lot more easier for the user.

jpa198:
> Just for my curiosity, what was the use case for this feature?

The current need is for Vahid to run IMs on large mag validation, which were generated using non uniform grid and so have way more stations than he needs

PR #122, opened 2021-08-09
Fix:Compute SDI via calculate_ims.py with -a Burks_Baker_2013

sungeunbae:
If -a Burks_Baker_2013 option is used with calculate_ims.py, it failed due to removal of "ver" component argument for run.py 
https://github.com/ucgmsim/IM_calculation/commit/022a9a1800d772da09c1db227cabaeb15f5d3824

Added required_components filed to advanced_im_config.yaml, so that a different set of components can be used for a specific IM.

Also added a hack to pass --rotd flag to run.py. For now, SDI is the only one that needs special care - will need to devise a better solution if more IMs need to be looked after

### DIFF
--- a/IM_calculation/scripts/calculate_ims.py
+++ b/IM_calculation/scripts/calculate_ims.py
@@ -222,6 +222,8 @@ def main():
     if args.advanced_ims is not None:
         components = advanced_IM_factory.COMP_DICT.keys()
         extra_flags = []
+        # TODO: this could utilize check_rotd(), but adds extra complexity handling enums..
+        # TODO: more elegant solution could be desired if more IMs need special care
         if "Burks_Baker_2013" in args.advanced_ims and set(
             ["rotd50", "rotd100", "rotd100_50"]
         ).intersection(args.components):
--- a/IM_calculation/scripts/calculate_ims.py
+++ b/IM_calculation/scripts/calculate_ims.py
@@ -220,13 +220,16 @@ def main():
     station_names = args.station_names
     if args.advanced_ims is not None:
         components = advanced_IM_factory.COMP_DICT.keys()
-        extra_flags = []
+        extra_flags = {}
         # TODO: this could utilize check_rotd(), but adds extra complexity handling enums..
         # TODO: more elegant solution could be desired if more IMs need special care
-        if "Burks_Baker_2013" in args.advanced_ims and set(
-            ["rotd50", "rotd100", "rotd100_50"]
-        ).intersection(args.components):
-            extra_flags.append("--rotd")
+        for adv_im in args.advanced_ims:
+            extra_flags[adv_im] = []
+            if adv_im == "Burks_Baker_2013":
+                if set(["rotd50", "rotd100", "rotd100_50"]).intersection(
+                    args.components
+                ):
+                    extra_flags[adv_im].append("--rotd")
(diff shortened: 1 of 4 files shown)


IM_calculation/scripts/calculate_ims.py
```diff
@@ -220,9 +220,17 @@ def main():
...
+        # TODO: more elegant solution could be desired if more IMs need special care
+        if "Burks_Baker_2013" in args.advanced_ims and set(
+            ["rotd50", "rotd100", "rotd100_50"]
+        ).intersection(args.components):
+            extra_flags.append("--rotd")
         advanced_im_config = advanced_IM_factory.advanced_im_config(
-            args.advanced_ims, args.advanced_im_config, args.OpenSees_path
+            args.advanced_ims, args.advanced_im_config, args.OpenSees_path, extra_flags
         )
+
```
jpa198, line 233: Adding --rotd to extra flag means that it will be added to the call of every model, not just Burks_Baker. So if two models are given and rotd_50 is specified, this will crash.
I think that if you want rotd_50 you should only be running with burks_baker, or you run without rotd_50, and throw an exception if they try to do both

jpa198 (changes requested):

IM_calculation/scripts/calculate_ims.py
sungeunbae: oh, that's a good point.

sungeunbae: @jpa198  Hopefully you like my new solution better.

jpa198 (approved):

jpa198 (approved):

IM_calculation/scripts/calculate_ims.py
jasonmotha: I think that rotd_50 flag should be set as part of the implementation of the adv_im or if you actually do need to turn it on / off advanced im_config seems like the more appropriate place to store it.

jasonmotha: Or configuration in the Model folder itself. Advanced IMs are made to be quite self contained, rather than having configuration passed through IM_calc. The reason for that is that each model is quite unique and has different configuration based on the IM.

jasonmotha:
Just for my curiousity, who did you plan this with? as it is structured change.

sungeunbae:
> Just for my curiousity, who did you plan this with? as it is structured change.

Discussed with @jpa198 - this is an issue introduced  by SDI_rotd PR and appeared to be high priority.

IM_calculation/scripts/calculate_ims.py
sungeunbae: We discussed it and decided against it 0 advanced im_config is not an ideal place to turn things on and off. At the moment, SDI is the only advanced IM that needs rotd calculated. We could certainly devise a more elegant solution later with more IMs in need of rotd coming along. Currently we have a use case where this issue was blocking and switching to run.py was not an ideal alternative (lack of multi-core support, orchestration of the workflow, support for binary waveform etc).

shot0829 (approved):

sungeunbae:
@jasonmotha  Can we merge this? This is a minor structural change that doesn't cause a degradation, and it doesn't pretend to be the perfect solution. For now, rotd is only required by SDI, and this is a reasonable solution considering the current structure and user interface We could think of a better solution when the time is ripe.

PR #124, opened 2021-08-12
Aggregates the status view of advanced IM for all events 

sungeunbae:
(python3_maui) baes@maui02: ~$ python overview_adv_IM_status.py /nesi/nobackup/nesi00213/RunFolder/vlo23/first_release_210623_ATC_Steel_SCBF_3Story ATC_Steel_SCBF_3Story sim
Screen Shot 2021-08-12 at 3 04 39 PM


IM_calculation/Advanced_IM/overview_adv_IM_status.py
```diff
@@ -0,0 +1,79 @@
...
+
+    args = parser.parse_args()
+
+    if not args.root_path.is_dir():
+        raise AssertionError("Provide a valid directory path")
+
+    if args.type == "sim":
+        data_path = args.root_path / SIM_DIR_SUFFIX.format(args.model)
+    else:
+        data_path = args.root_path / OBS_DIR_SUFFIX.format(args.model)
```
jasonmotha, line 77: Would a dictionary be a better way to do this rather than conditionals?

I.e. data_path = DIR_SUFFIX[args.type] ?

IM_calculation/Advanced_IM/overview_adv_IM_status.py
```diff
@@ -0,0 +1,79 @@
...
+
+    overall_status_df = pd.DataFrame.from_dict(
+        event_status_dict, orient="index", columns=["total"] + ALL_STATUS
+    )
+    overall_status_df.to_csv(outdir / f"overall_status_{model}_{type}.csv")
+    print(overall_status_df)
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser()
```
jasonmotha, line 57: Ideally put everything in here within a function so these variables / args don't end up in the global namespace.

IM_calculation/Advanced_IM/overview_adv_IM_status.py
```diff
@@ -0,0 +1,79 @@
...
+            except KeyError:
+                count = 0
+            event_status_dict[event_name][st] = count
+            event_status_dict[event_name]["total"] += count
+
+    overall_status_df = pd.DataFrame.from_dict(
+        event_status_dict, orient="index", columns=["total"] + ALL_STATUS
+    )
+    overall_status_df.to_csv(outdir / f"overall_status_{model}_{type}.csv")
+    print(overall_status_df)
```
jasonmotha, line 53: Is this print meant to be here?

IM_calculation/Advanced_IM/overview_adv_IM_status.py
```diff
@@ -0,0 +1,79 @@
+#!/usr/bin/env python
```
jasonmotha, line 1: This file (and ` check_adv_IM_status.py `) would be better in a scripts folder to differentiate between importable code and scripts.

My recommendations would be either IM_calc/scripts/adv_im/
or 
IM_calc/Adv_im/scripts/

I think I'm slightly more partial to the former.

IM_calculation/Advanced_IM/overview_adv_IM_status.py
```diff
@@ -0,0 +1,79 @@
...
+    "not_converged",
+    "not_finished",
+    "timed_out",
+    "crashed",
+    "unknown",
+]
+TYPES = ["obs", "sim"]
+
+SIM_DIR_SUFFIX = "Runs/*/*/IM_calc/{}_status.csv"
+OBS_DIR_SUFFIX = "ObservedGroundMotions/IM_calc/*/{}_status.csv"
```
jasonmotha, line 22: Because of the assumed knowledge about sim_structure this script might be better off in sim_workflow.
it's generally assumed that im_calc, empirical_engine, visualization etc. don't know the cs folder structure, so all the folder structure code is in one place. Especially around the retrieval of the fault/rel_name from the folder structure.

Perhaps a better solution would be to pass all the *_status.csv filenames as a nargs=+ to this script and then add it to the slurm script for advanced IMs?

Thoughts @jpa198 ?

IM_calculation/Advanced_IM/overview_adv_IM_status.py
```diff
@@ -0,0 +1,79 @@
...
+    overall_status_df.to_csv(outdir / f"overall_status_{model}_{type}.csv")
+    print(overall_status_df)
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser()
+    parser.add_argument(
+        "root_path",
+        type=Path,
+        help="Root path of GM sim.ie. Parent directory of Runs or ObservedGroundMotiohs",
```
jasonmotha, line 61: motions.

IM_calculation/Advanced_IM/overview_adv_IM_status.py
```diff
@@ -0,0 +1,79 @@
...
+    "unknown",
+]
+TYPES = ["obs", "sim"]
+
+SIM_DIR_SUFFIX = "Runs/*/*/IM_calc/{}_status.csv"
+OBS_DIR_SUFFIX = "ObservedGroundMotions/IM_calc/*/{}_status.csv"
+
+
+def process(type, model, path, outdir):
+    event_status_files = glob.glob(str(path))
```
jasonmotha, line 26: If you are using Path objects, there is an in-built glob operator. path.glob(match_string)

IM_calculation/Advanced_IM/overview_adv_IM_status.py
```diff
@@ -0,0 +1,79 @@
...
+
+    event_status_dict = {}
+    for event_status_file in event_status_files:
+        if type == "sim":
+            event_name = Path(event_status_file).parents[2].name
+        else:
+            event_name = Path(event_status_file).parent.name
+        event_status_df = pd.read_csv(event_status_file)
+        event_status_dict[event_name] = {"total": 0}
+        for st in ALL_STATUS:
```
jasonmotha, line 41: I'd use the full name `status`, just to be clear what it is.

jasonmotha (changes requested):

IM_calculation/Advanced_IM/overview_adv_IM_status.py
```diff
@@ -0,0 +1,79 @@
+#!/usr/bin/env python
+# coding: utf-8
+
+import argparse
+import glob
+import pandas as pd
+from pathlib import Path
+import sys
+
+ALL_STATUS = [
```
shot0829, line 10: why not import the class from the check_status script? to make sure they stay the same.

IM_calculation/Advanced_IM/overview_adv_IM_status.py
shot0829: my guess is its intentional. As it would show the user a overall status without the need to open the csv file.

jasonmotha: It's the whole dataframe? I guess it won't be many lines for smaller runs? But for small mag, it would be like 500 lines?

IM_calculation/Advanced_IM/overview_adv_IM_status.py
jpa198: I'm definitely against having to pass in the glob strings manually every time, when they're basically the same every time. This isn't an event/realisation level script, it's a sim root level script, so wouldn't fit into any automated workflow slurm script. (It could, but it would result in realisations interacting with each other which has been a no-go so far)
Having a third 'TYPES' entry for custom could be good if we have a scenario where the output folder structure doesn't match the current sim or obs structures.
Sending this script over to get lost in the scripts folder their seems like overkill too.

I think this script is fine where it is, if you add a custom TYPE and consider that the default, and the sim/obs to be commonly used shortcuts.

IM_calculation/Advanced_IM/overview_adv_IM_status.py
```diff
@@ -0,0 +1,79 @@
...
+            event_name = Path(event_status_file).parents[2].name
+        else:
+            event_name = Path(event_status_file).parent.name
+        event_status_df = pd.read_csv(event_status_file)
+        event_status_dict[event_name] = {"total": 0}
+        for st in ALL_STATUS:
+            try:
+                count = event_status_df["status"].value_counts()[st]
+            except KeyError:
+                count = 0
```
jpa198, line 45: I've done a quick test with value_counts, and it looks like it accepts dict get() notation. So you should be able to do this:`event_status_df["status"].value_counts().get(st, 0)` instead of having the try/except

IM_calculation/Advanced_IM/overview_adv_IM_status.py
jpa198: Perhaps something like describe() would be more useful here. With the count for each column total, so if nothing's wrong you don't even need to look in the output file.

IM_calculation/Advanced_IM/overview_adv_IM_status.py
sungeunbae: Good idea

IM_calculation/Advanced_IM/overview_adv_IM_status.py
sungeunbae: Sure - agreed

IM_calculation/Advanced_IM/overview_adv_IM_status.py
sungeunbae: I'm with @jpa198 and don't think this belongs to sim_workflow. This is specific to advanced IM.  A custom TYPE is an interesting idea, but it will take a bit of work to accommodate a random directory structure. Will get it implemented shortly

IM_calculation/Advanced_IM/overview_adv_IM_status.py
sungeunbae: sure

IM_calculation/Advanced_IM/overview_adv_IM_status.py
sungeunbae: Yes, but I am not sure if it makes the code any simpler. path.glob returns a generator and can't use len().

IM_calculation/Advanced_IM/overview_adv_IM_status.py
sungeunbae: This "print" is intentional - it also produces a CSV file and it will be more useful if there are lots of lines, but for a large/moderate mag, it saves opening a CSV file - the user requested this feature downloads CSV file and opens in Excel. Sure - Will add some fancy summary

IM_calculation/Advanced_IM/overview_adv_IM_status.py
sungeunbae: good idea

IM_calculation/Advanced_IM/overview_adv_IM_status.py
jasonmotha: I think in terms of simplicity, it's slightly more simple as it avoids the conversion from a path into a string and then back into a path later. But i do see your point.

The len(generator) is very interesting, though it does make sense that it doesn't exist. I guess a lot of time that I've used generators I've used list(generator_expression) which kinda is against the point of generators.

jpa198: list(path.glob()) has less type conversion than Path(x) where x is in glob.glob(str(path)), so I'm in favour of it, even if it does defeat the purpose of generators

IM_calculation/Advanced_IM/overview_adv_IM_status.py
sungeunbae: When this is a stand-alone script, does the global namespace matter?

jpa198: If we need to change it later, making sure we don't have to worry about things in the global namespace will make it easier.
Just yesterday I was cursing the global namespace in plot_items

IM_calculation/Advanced_IM/overview_adv_IM_status.py
sungeunbae: @jpa198 
Perhaps this is not what you want to see.
```
>>> df.describe().loc["count"]
total            14.0
finished         14.0
not_started      14.0
not_converged    14.0
not_finished     14.0
timed_out        14.0
crashed          14.0
unknown          14.0
```
My fancy output is this:
```
(python3_maui) baes@maui02: ~$ python overview_adv_IM_status.py /nesi/nobackup/nesi00213/RunFolder/vlo23/first_release_210623_ATC_Steel_SCBF_3Story ATC_Steel_SCBF_3Story sim
-------
1 incomplete/unsuccessful event(s) found (Total: 14)
             total  finished  not_started  not_converged  not_finished  timed_out  crashed  unknown
2016p858000    302       301            0              1             0          0        0        0
(python3_maui) baes@maui02: ~$ python overview_adv_IM_status.py /nesi/nobackup/nesi00213/RunFolder/vlo23/first_release_210623_ATC_Steel_SCBF_3Story ATC_Steel_SCBF_3Story obs
-------
All completed successfully! (Total: 14)
```
I actually considering removing CSV output altogether. I think this terminal output should be good enough

IM_calculation/Advanced_IM/overview_adv_IM_status.py
jasonmotha: I'd say that a custom type is a good idea, but it's probably not worth the development time for it currently. If there's a need we can implement it later.

If at all possible, can we abstract the sim_structure assumptions to qcore.sim_structure?

IM_calculation/Advanced_IM/overview_adv_IM_status.py
jasonmotha: I'm probably missing something from that output above how does a user tell which runs need attention?

sungeunbae: **2016p858000**

IM_calculation/Advanced_IM/overview_adv_IM_status.py
sungeunbae: Sure - It needs a bit more code change than you might be suggesting (`path` here is a Path object, but as it has an embedded pattern, it can't run glob by itself) - anyway I got it working. Had to use list().

IM_calculation/Advanced_IM/overview_adv_IM_status.py
sungeunbae: calling main() function from __main__

IM_calculation/Advanced_IM/overview_adv_IM_status.py
jpa198: For a script that doesn't even import qcore atm, that seems a bit over kill

sungeunbae: I agree - and the pattern is not too difficult to figure out when needed

jpa198: Would that even work? It'd have to be pretty jank

IM_calculation/Advanced_IM/overview_adv_IM_status.py
jasonmotha: Ahhhh, I thought that was the name of the run rather than the event. If there are multiple do they appear in a list?

sungeunbae: Yes, each incomplete, unsuccessful event per each row

IM_calculation/Advanced_IM/overview_adv_IM_status.py
jpa198: Ok, if you give the realisation as '*' it would spit out the same, so it would work. Yeah, it probably is safest to import sim_struct instead of leaving the hardcoded in there, in case we change the locations in future or something

IM_calculation/Advanced_IM/overview_adv_IM_status.py
jpa198: So if each event in small mag has an issue, the user will still get bombarded with 500 lines of issues in their terminal?

jpa198: That first image from the three above is what I was pushing for. Just something to say something went wrong vs everything's fine, depending on the status counts. Then the user can read the full file to find the issues without their terminal getting swamped

sungeunbae: yes, and it means the run has some serious issue. `|head N` or `|tail N` can help

jasonmotha: That's a good point. Seems like the count of each status type is a good output summary and then per-realisation information in the CSV is the next debug step.

sungeunbae: First of all, let see what James' preferred summary is about.
```
>>> df.describe().loc["count"]
total            14.0
finished         14.0
not_started      14.0
not_converged    14.0
not_finished     14.0
timed_out        14.0
crashed          14.0
unknown          14.0
```
There are 14 events, hence 14 records for each column, which gives no value to the user. Perhaps you are referring to this?
```
>>> df.describe().loc['max']
total            302.0
finished         301.0
not_started        0.0
not_converged      1.0
not_finished       0.0
timed_out          0.0
crashed            0.0
unknown            0.0
Name: max, dtype: float64
```

BTW, this is the full df
```
>>> df
             total  finished  not_started  not_converged  not_finished  timed_out  crashed  unknown
2103645         14        14            0              0             0          0        0        0
2013p543824     86        86            0              0             0          0        0        0
3468575         48        48            0              0             0          0        0        0
3366146        108       108            0              0             0          0        0        0
2016p858000    302       301            0              1             0          0        0        0
2839343         10        10            0              0             0          0        0        0
3124785         34        34            0              0             0          0        0        0
2014p051675     22        22            0              0             0          0        0        0
3631380         58        58            0              0             0          0        0        0
2013p613797     94        94            0              0             0          0        0        0
2015p012816      8         8            0              0             0          0        0        0
3528839         40        40            0              0             0          0        0        0
2808298         18        18            0              0             0          0        0        0
2016p661332     14        14            0              0             0          0        0        0
```

The user sees that there is 1 non_converged case, but the user doesn't know which event it is. So the user has to open the overview CSV, and find it is 2016p858000. Then goes to the event's own CSV for further debug.

I'm thinking of removing the overview CSV altogether. 
If there are many lines and the screen is swamped, it means the whole thing is screwed or it is too early to have majority of runs in "finished" state.  `head` or `tail` commands can come in handy. 
The debug step makes sense when there are only *a few* unsuccessful/incomplete events, and having the directly relevant info on the screen saves the extra step. 

If you have seen the user (who requested this feature) downloading the CSV file from mobaxterm and opens it with Excel, I will find it easier to convince you that my way of "summary" works better.

jasonmotha: Our concern is that it works for a small number of events. But if you are running it with small mag ~500 events, that becomes unweildly on a terminal. 

I initially thought total was for 14 events total, rather than 14 stations within the event. 

I think a summary of the entire run is most useful. Most of the time we want an overview summary and then ability to go into things if the need arises.

sungeunbae: ok - I can compromise. By default, it will just print the summary and store the extra info in csv. With extra switch --screen (default false), it will give the summary I like.

jasonmotha: I do like the idea of having a verbose optional output.

I think it's standard to call it `-v`, `--verbose` however.

jpa198: Ah, my mistake, I thought the "count" would give the sum of each column for some reason. Having the sum in the output would be most useful as you can then go through the csv ticking them off as you go. 
Having the full summary is fine for a handful of events, but definitely not small magnitude (let alone using perturbations). Also having the file lets you go back and check if you need to, vs the transient nature of terminal display

IM_calculation/Advanced_IM/overview_adv_IM_status.py
sungeunbae: As in path.glob(pattern), something that has "*" should be pattern, not path, I had to make a bit of rearrangement. This should make you both happy.

IM_calculation/Advanced_IM/overview_adv_IM_status.py
sungeunbae: Gone into IM_calculation/IM_calculation/Advanced_IM/scripts

IM_calculation/Advanced_IM/overview_adv_IM_status.py
sungeunbae: --verbose added. Writes to csv as it used to and prints only
```
------------
total            856
finished         855
not_started        0
not_converged      1
not_finished       0
timed_out          0
crashed            0
unknown            0
------------
1 incomplete/unsuccessful event(s) found (Total: 14)
```
but no event info unless --verbose is specified.

IM_calculation/Advanced_IM/scripts/overview_adv_IM_status.py
```diff
@@ -0,0 +1,135 @@
+#!/usr/bin/env python
+# coding: utf-8
+
+import argparse
+import glob
+import os.path
+import pandas as pd
```
jasonmotha, line 7: pandas isn't built in so should be in a separate group.

IM_calculation/Advanced_IM/overview_adv_IM_status.py
sungeunbae: custom type was also implemented

IM_calculation/Advanced_IM/scripts/overview_adv_IM_status.py
```diff
@@ -0,0 +1,135 @@
...
+from pathlib import Path
```
jasonmotha, line 8: Correct me if I'm wrong but pep nomally puts from at the start and then the other imports alphabetically.

IM_calculation/Advanced_IM/scripts/overview_adv_IM_status.py
```diff
@@ -0,0 +1,135 @@
...
+    event_status_dict = {}
+    for event_status_file in event_status_files:
+        if type == "sim":
+            event_name = Path(event_status_file).parents[2].name
+        else:
+            event_name = Path(event_status_file).parent.name
+        event_status_df = pd.read_csv(event_status_file)
+        event_status_dict[event_name] = {"total": 0}
+        for status in ALL_STATUS:
+
```
jasonmotha, line 50: Is this whitespacing intentional?

IM_calculation/Advanced_IM/scripts/overview_adv_IM_status.py
```diff
@@ -0,0 +1,135 @@
...
+        "-v",
+        "--verbose",
+        action="store_true",
+        default=False,
+        help="Print unfinished events summary",
+    )
+
+    args = parser.parse_args()
+
+    if not args.sim_root.is_dir():
```
jasonmotha, line 113: Would a direct assert statement be better instead of raising an assert if [conditional]?

IM_calculation/Advanced_IM/scripts/overview_adv_IM_status.py
```diff
@@ -0,0 +1,135 @@
...
+            args.pattern = args.pattern.strip(
+                "/"
+            )  # remove leading or trailing "/" (if any)
+    else:
+        if args.pattern is not None:
+            print(
+                f"Warning: Type {args.type} uses pre-defined search pattern - supplied one ignored"
+            )
+
+    process(
```
jasonmotha, line 129: Would aggregate_status_files be a better function name?

jasonmotha (changes requested):
Lots of big changes. Starting to look much better! =D

IM_calculation/Advanced_IM/scripts/overview_adv_IM_status.py
sungeunbae: Perhaps less helpful to user? 
```
Traceback (most recent call last):
  File "overview_adv_IM_status.py", line 137, in <module>
    main()
  File "overview_adv_IM_status.py", line 116, in main
    assert args.sim_root.is_dir()
AssertionError

```

IM_calculation/Advanced_IM/scripts/overview_adv_IM_status.py
sungeunbae: Sure - to be changed in the next commit

IM_calculation/Advanced_IM/scripts/overview_adv_IM_status.py
jasonmotha: `assert args.sim_root.is_dir(), "Provide a valid directory path"` ?

Just don't put brackets around it because tuples are True. I've made that mistake before.

jpa198 (dismissed):
Cool, just those couple of things from Jason, but looks good to me

IM_calculation/Advanced_IM/scripts/overview_adv_IM_status.py
sungeunbae: ok

IM_calculation/Advanced_IM/scripts/overview_adv_IM_status.py
sungeunbae: my mind must have been blank

IM_calculation/Advanced_IM/scripts/overview_adv_IM_status.py
sungeunbae: Done

IM_calculation/Advanced_IM/scripts/overview_adv_IM_status.py
sungeunbae: Done

IM_calculation/Advanced_IM/scripts/overview_adv_IM_status.py
sungeunbae: Done

shot0829 (approved):
make sure the script location is update in 
https://github.com/ucgmsim/slurm_gm_workflow/blob/186bd14f68b4991a34cd2606fec36993c08a7f52/templates/adv_im_calc.sl.template#L36
and
https://github.com/ucgmsim/slurm_gm_workflow/blob/master/scripts/nesi_scripts/run_adv_im_obs_maui.sl

jpa198 (approved):

jasonmotha (commented):
image
Pretty happy on my comments. 

Though Jonney's one seems to have dropped off the face of the earth. 

I wouldn't import from a script though... But should be pretty simple to drum up a class?

sungeunbae:
> image
> Pretty happy on my comments.
> 
> Though Jonney's one seems to have dropped off the face of the earth.
> 
> I wouldn't import from a script though... But should be pretty simple to drum up a class?

My literacy level finds this comment incomprehensible. Where is the comment in the screenshot from, and is this Jonney's or yours'? Could you rephrase in plain English?

sungeunbae:
> > image
> > Pretty happy on my comments.
> > Though Jonney's one seems to have dropped off the face of the earth.
> > I wouldn't import from a script though... But should be pretty simple to drum up a class?
> 
> My literacy level finds this comment incomprehensible. Where is the comment in the screenshot from, and is this Jonney's or yours'? Could you rephrase in plain English?

OK - I think I should expand my reading list - never been a big fan of poetic works. Guys in my room help me and I can see a better picture now. 

Jonney removed his comments which I have never seen,  and the first sentence gave me a little wonder why you are being happy with your own comments.  

A question remains. I'm not sure how I can drum up a class when there is already a class, while I must admit, it sounds quite exciting.

sungeunbae:
> > > image
> > > Pretty happy on my comments.
> > > Though Jonney's one seems to have dropped off the face of the earth.
> > > I wouldn't import from a script though... But should be pretty simple to drum up a class?
> > 
> > 
> > My literacy level finds this comment incomprehensible. Where is the comment in the screenshot from, and is this Jonney's or yours'? Could you rephrase in plain English?
> 
> OK - I think I should expand my reading list - never been a big fan of poetic works. Guys in my room help me and I can see a better picture now.
> 
> Jonney removed his comments which I have never seen, and the first sentence gave me a little wonder why you are being happy with your own comments.
> 
> A question remains. I'm not sure how I can drum up a class when there is already a class, while I must admit, it sounds quite exciting.

OK - the updated interpreter now understands the last question. Working on the change now

jasonmotha (dismissed):

sungeunbae:
All 3 reviewers approved - had to make one last commit to get Jenkins black test passed. Enforcing merge now.

PR #126, opened 2021-08-18
Advanced SDI with calculate_ims.py

sungeunbae:
calculate_ims.py -a option will conflict with -c. If -a is specified, it will return all components run.py produces.
As a result, SDI if computed with calculated_ims.py -a option, will compute rotd by default.

SDI/run.py will skip rotd if norotd is set.

### DIFF
--- a/IM_calculation/Advanced_IM/advanced_IM_factory.py
+++ b/IM_calculation/Advanced_IM/advanced_IM_factory.py
@@ -67,7 +67,7 @@ def compute_ims(accelerations, configuration, adv_im_out_dir):
                     ["--timeout_threshold", str(im_config["timeout_threshold"])]
                 )
             # add im specific flags
-            script.extend(configuration.extra_flags)
+            script.extend(configuration.extra_flags[im])
             print(" ".join(script))
             subprocess.run(script)
 
--- a/IM_calculation/Advanced_IM/advanced_IM_factory.py
+++ b/IM_calculation/Advanced_IM/advanced_IM_factory.py
@@ -14,7 +14,7 @@ advanced_im_dir = os.path.dirname(__file__)
 CONFIG_FILE_NAME = os.path.join(advanced_im_dir, "advanced_im_config.yaml")
 
 advanced_im_config = namedtuple(
-    "advanced_im_config", ["IM_list", "config_file", "OpenSees_path", "extra_flags"]
+    "advanced_im_config", ["IM_list", "config_file", "OpenSees_path"]
 )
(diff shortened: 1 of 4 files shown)


IM_calculation/Advanced_IM/advanced_IM_factory.py
```diff
@@ -54,6 +54,9 @@ def compute_ims(accelerations, configuration, adv_im_out_dir):
             ]
             # waveform component sequence
             comp_list = ["000", "090", "ver"]
+            if "required_components" in im_config:
+                comp_list = im_config["required_components"]
```
jpa198, line 58: Could be 
comp_list = im_config.get("required_components", ["000", "090", "ver"])

jpa198 (changes requested):
Did we agree to not go with the plan of "allow SDI to take in ver optionally, so that it won't crash if we gave it ver anyway"? I don't remember the outcome of that conversation

IM_calculation/Advanced_IM/advanced_IM_factory.py
sungeunbae: I can't quite remember either - I thought optional ver component could work, but looking at the code again, I can't think of an elegant solution for that. Perhaps requirement_components can be taken from advanced_im_config.yaml into runlibs_2d.py as we do in advanced_IM_factory.py and we could make parse_args() more intelligent. 
Other than that, I really don't see why we should enable ver component for SDI

sungeunbae: In fact, we don't need --OpenSees_path for SDI either, but  it is added regardless. It does no harm, but pointless nonetheless

jasonmotha (commented):
I can't remember hearing about this task as part of the sprint planning?
I thought the only advanced IM change for this sprint was adding new models?

shot0829:
> I can't remember hearing about this task as part of the sprint planning?
> I thought the only advanced IM change for this sprint was adding new models?

if i remember correctly, this was a closed PR that at the time we thought might not be needed.
And this was revived because there was a usage from vahid i think, maybe sung can give you more detail.

sungeunbae:
> > I can't remember hearing about this task as part of the sprint planning?
> > I thought the only advanced IM change for this sprint was adding new models?
> 
> if i remember correctly, this was a closed PR that at the time we thought might not be needed.
> And this was revived because there was a usage from vahid i think, maybe sung can give you more detail.

We closed but discussed what to do if it causes a pain to manage two separate branches. Last week, I asked James to add this to Sprint planning, but has been forgotten.

PR #130, opened 2021-08-25
Added 3 new adv IM models

sungeunbae:


shot0829 (approved):
looks good.
As mentioned in slack, if the test on large mag waveforms pass then its good to go.

PR #135, opened 2021-09-01
PG\* rotd50 was taking the median etc off the wrong axis.

jasonmotha:
Answers before:
image
Answers after:
image

Unsure if other IMs such as SDI are affected (but because it's multi-dimensional; it's probably okay like pSA)

### DIFF
--- a/IM_calculation/IM/im_calculation.py
+++ b/IM_calculation/IM/im_calculation.py
@@ -100,10 +100,17 @@ def check_rotd(comps_to_store: Iterable[Components]) -> bool:
 def calculate_rotd(
     accelerations,
     comps_to_store: List[Components],
-    func=lambda x: np.max(np.abs(x), axis=1),
+    func=lambda x: np.max(np.abs(x), axis=-2),
 ):
     """
     Calculates rotd for given accelerations
+
+    Multiplying the waveforms by the rotation matrices gives an array with shape either
+    (periods, nt, n_rotations) (for pSA) or (nt, n_rotations) (for single dimension IMs).
+    The max is taken over the second axis from the rear (nt),
+    so for pSA we get an array with shape (periods, n_rotations), while single dimension IMs gets (n_rotations)
+    Which is then taken the median for RotD50 and the maximum for RotD100
+
     :param accelerations: An array with shape [[periods.size,] nt, 2]
         where the first axis is optional and if present is equal to the number of periods in the intensity measure.
         nt is the number of timesteps in the original waveform
@@ -228,7 +235,7 @@ def calc_PG(waveform, im, comps_to_store, comps_to_calculate):
     value = intensity_measures.get_max_nd(waveform)
     values = array_to_dict(value, comps_to_calculate, im, comps_to_store)
     if check_rotd(comps_to_store):
-        rotd = calculate_rotd(waveform, comps_to_store, func=lambda x: np.max(np.abs(x), axis=0))
+        rotd = calculate_rotd(waveform, comps_to_store)
         sanitise_single_value_arrays(rotd)
         values.update(rotd)
     return values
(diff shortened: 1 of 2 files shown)


lee-robin (dismissed):
I guess this means some other discrete IMs like AI and CAV will have this issue. Ds may not though if it's an array of Ds575 and  Ds595?

Edit: Also, what was it actually taken the median of when the axis is wrong? Of all records? (i.e. when correct it's the median across all the rotation angles)

Edit2: Actually it can't be records, it would have to be median over IMs being calculated? Which would just be PGA? I can't make sense of it.

jpa198:
> I guess this means some other discrete IMs like AI and CAV will have this issue. Ds may not though if it's an array of Ds575 and Ds595?
> 
> Edit: Also, what was it actually taken the median of when the axis is wrong? Of all records? (i.e. when correct it's the median across all the rotation angles)

So multiplying the waveforms by the rotation matrices gives an array with shape either (periods, nt, n_rotations) (for pSA) or (nt, n_rotations) (for PGA, PGV). We take the max over the second axis, so for pSA we get an array with shape (periods, n_rotations), while PG* gets (nt), so for PGV at each timestep it was taking the maximum across rotations, then the middle/max value was taken from them for rotd50/100.

Changing the default function to use axis -2 would be a more universal way of solving this I think.
AI and CAV have custom functions, so don't suffer from this issue

This also indicates any PGV/PGA rotd calculations we've ever done are trash.

jpa198:
Currently for rotd50 we're doing this (with array shapes in brackets):
pSA:
`np.median(np.max(np.dot((n_p, nt, xy), (xy, n_theta)), axis=1), axis=-1) = np.median(np.max((n_p, nt, n_theta), axis=1), axis=-1)
                                                                                                   = np.median((n_p, n_theta), axis=-1)
                                                                                                   = (n_p)`
current PG*:
`np.median(np.max(np.dot((nt, xy), (xy, n_theta)), axis=1), axis=-1) = np.median(np.max((nt, n_theta), axis=1), axis=-1)
                                                                                                         = np.median((nt), axis=-1)
                                                                                                          =(1)`
proposed PG*:
`np.median(np.max(np.dot((nt, xy), (xy, n_theta)), axis=0), axis=-1) = np.median(np.max((nt, n_theta), axis=0), axis=-1)
                                                                                                         = np.median((n_theta), axis=-1)
                                                                                                          =(1)`

jasonmotha:
PG* being different makes these plots check out https://wiki.canterbury.ac.nz/pages/viewpage.action?pageId=110430202

IM_calculation/IM/im_calculation.py
```diff
@@ -100,10 +100,17 @@ def check_rotd(comps_to_store: Iterable[Components]) -> bool:
 def calculate_rotd(
     accelerations,
     comps_to_store: List[Components],
-    func=lambda x: np.max(np.abs(x), axis=1),
+    func=lambda x: np.max(np.abs(x), axis=-2),
```
jpa198, line 103: The default function should also be changed in intensity_measures.get_rotations

jasonmotha: Ahhh, helps if you add all the files you changed to your commit.

jpa198: lol

lee-robin (approved):
Thanks for the explanations, makes more sense what was going on now. Hoping this -2 solution will be relatively robust going forward? Only trouble when going to even higher dimensions potentially?

jasonmotha:
Yeah, should only be an issue for other new functions that want to use the rotd functionality. Did we test RotD values against seismosignal or something?

SquirrelKnight:
Elena noticed this issue with her H/V computations. After investigating the
outputs, I found that the PGA values for rotd50 were significantly lower
than 000, 090, ver, etc.

PR #136, opened 2021-09-02
Creates virtual env

sungeunbae:
Testing


Jenkinsfile
```diff
@@ -43,8 +40,8 @@ pipeline {
                     echo "[ Python used ] : " `which python`
                     cd ${env.WORKSPACE}
                     echo "[ Installing ${env.JOB_NAME} ]"
-# full installation is not possible as it takes more than 3.0Gb for building
-#                    python setup.py install
+# full installation is not possible as it takes more than 3.0Gb for building and kills the server
+#                   python setup.py install
                     python setup.py build_ext --inplace
 		            python konno_setup.py
```
jasonmotha, line 46: That's odd that you can create the konno matrices manually but not from the setup.py script?

requirements.txt
```diff
@@ -1,9 +1,10 @@
 pytest>=3.5.0
 pytest-black
-numpy
+numpy==1.20
```
jasonmotha, line 3: If you are using isclose do you have to hardcode the numpy version?

sungeunbae: Good catch

Jenkinsfile
sungeunbae: konno_setup.py has a limited depth (num_to_gen=5), just enough for auto testing. 
Konno matrices built by setup.y runs in at memory alloc failure (2.0Gib) at 16385

requirements.txt
sungeunbae: BTW, fixing numpy==1.20 was not good enough to produce the result giving the same number of significant digits. Finding out what exact package was causing this was going to be a treasure hunt promising little prize..

jasonmotha (approved):

jasonmotha (approved):

sungeunbae:
Had to re-introduce numpy==1.20 as numba doesn't like the latest numpy

jpa198 (approved):

jpa198 (approved):

jasonmotha (approved):

PR #137, opened 2021-09-09
forced the version of numba, to solve a llvmlite crash issue on nurion

shot0829:


jasonmotha (changes requested):
I'm also not sure that this place is the best to fix this issue for one platform as it influences all the platforms?

I think we still need numpy on the list?
It would be good to have a '>=' on numba so that it isn't limited to that version until we notice it again.

shot0829:
> I'm also not sure that this place is the best to fix this issue for one platform as it influences all the platforms?
> 
> I think we still need numpy on the list?
> It would be good to have a '>=' on numba so that it isn't limited to that version until we notice it again.

i removed numpy because it is required for numba.

shot0829:
will it be better if i use >= instead of ==?
the issue was for some reason pip's default version is not the latest and installed a old version which caused all kinds of trouble.

shot0829:
nevermind, i missed the line in your comment that suggested >=

sungeunbae:
> > I'm also not sure that this place is the best to fix this issue for one platform as it influences all the platforms?
> > I think we still need numpy on the list?
> > It would be good to have a '>=' on numba so that it isn't limited to that version until we notice it again.
> 
> i removed numpy because it is required for numba.

As numba was picky with the version of numpy, removing numpy and letting numba decide the version of numpy appears to be a clever workaround to me.

jpa198 (approved):
I don't think setting a minimum level is a bad thing

sungeunbae (approved):

jasonmotha (approved):

PR #138, opened 2021-09-09
Adding gc2

jpa198:


IM_calculation/source_site_dist/src_site_dist.py
```diff
@@ -114,27 +144,74 @@ def calc_rx_ry(srf_points: np.ndarray, plane_infos: List[Dict], locations: np.nd
...
+    origin_offset = 0
+    if hypocentre_origin:
+        # Our faults only use one hypocentre
+        # Will only use the first one found if there are multiple
+        for plane in plane_infos:
+            if plane["shyp"] == -999.9000:
+                origin_offset -= plane["length"]
+            else:
+                origin_offset -= plane["length"] / 2 + plane["shyp"]
+                break
```
jasonmotha, line 194: This isn't my favourite block of logic. But I can't think of a better way to to it.

jasonmotha (dismissed):
Nice recycling.

Looks nice and clean too.

IM_calculation/source_site_dist/src_site_dist.py
jpa198: It would be perfect for a do while loop :P

IM_calculation/source_site_dist/src_site_dist.py
```diff
@@ -203,6 +203,9 @@ def calc_rx_ry_GC2(
                 plane_points, [plane_header], np.asarray([loc])
             )
             dists = h_dist_f(plane_points, loc[0], loc[1])
+            # Mimimum distance of 1m to prevent nans/infs
```
jasonmotha, line 206: update the comment to 0.001m or just remove the distance specifier.

jasonmotha (approved):

IM_calculation/source_site_dist/src_site_dist.py
jpa198: All distances used in geo and src site dist are in km. 0.001km == 1m

shot0829: the unit matters, i guess thats what jason's comment is about.

shot0829 (dismissed):

joelridden (approved):
Nice to see the hypocentre_origin flag, will come in handy.

IM_calculation/source_site_dist/src_site_dist.py
jasonmotha: Oh of course. It's confusing as is with having one number in one unit and just below it's in a different unit.

jpa198: @jasonmotha I'm not sure you understand the code + comment? The code distance is 0.001km, while the comment says 1m. So same distance in different units.

shot0829: well, yes it the same distance, but if I'm debugging and see '1', should i expect it to be 1m or 1km?
so if the value are in km, then the comment should also be in km, or which ever it is as long as it matches.

jasonmotha: Yeah - what Jonney said.

shot0829 (approved):

IM_calculation/source_site_dist/src_site_dist.py
jpa198: Oh right, github didn't show me your second comment before I posted mine Jason

jasonmotha (approved):

PR #141, opened 2021-09-17
Summary updates

jasonmotha:
Results look to be the same when changed to the vectorised version (easier to do than have another for loop or component specification)

Left is vectorised, middle is selecting only geom for the summary and right is old functionality

image

### DIFF
--- a/scripts/summarise_all_rels_im.py
+++ b/scripts/summarise_all_rels_im.py
@@ -88,7 +88,7 @@ if __name__ == "__main__":
             if im_types:
                 assert set(df.columns).issuperset(im_types), (
                     f"The following IMs aren't present in the IM csv: {', '.join(set(im_types).difference(df.columns))}"
-                    f"Available IMs: {', '.join(df.columns)}"
+                    f"\nAvailable IMs: {', '.join(df.columns)}"
                 )
                 rel_im_dfs.append(df[im_types])
             else:
@@ -105,8 +105,10 @@ if __name__ == "__main__":
         merged_im_df = pd.concat(rel_im_dfs, axis=0, keys=range(len(rel_im_dfs)))
         if stat_fn == "mean":
             log_mean_im = np.exp(np.log(merged_im_df).mean(level=[1, 2]))
-        else:
+        elif stat_fn == "median":
             log_mean_im = np.exp(np.log(merged_im_df).median(level=[1, 2]))
+        else:
+            raise Exception("Invalid average type set")
         log_stdev_im = np.log(merged_im_df).std(level=[1, 2], ddof=0)
 
         log_stdev_im.columns = [f"{im}_sigma" for im in log_stdev_im.columns]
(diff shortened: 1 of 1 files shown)


joelridden (approved):

scripts/summarise_all_rels_im.py
```diff
@@ -73,60 +78,42 @@
...
 
         rel_im_dfs = []
         for c in im_csv_paths:
-            df = pd.read_csv(c, index_col=0)
-            rel_im_dfs.append(df)
+            df = load_im_file_pd(c)
+            rel_im_dfs.append(df[im_types])
 
-        stations = list(set(rel_im_dfs[0].index))
+        stations = list(set(rel_im_dfs[0].index.get_level_values(0)))
```
jpa198, line 92: remove list(set()) add .unique()

scripts/summarise_all_rels_im.py
```diff
@@ -73,60 +78,42 @@
...
         stations.sort()
-
         # check IM types. If unspecified, use all IM_types
         wrong_im_count = 0
```
sungeunbae, line 95: Looks like this variable is no longer needed?

sungeunbae (changes requested):

scripts/summarise_all_rels_im.py
```diff
@@ -25,35 +29,36 @@
...
+        const="median",
+        default="mean",
         help="find median (default: mean)",
     )
     parser.add_argument(
         "--im",
         dest="im_types",
         nargs="+",
         type=str,
+        choices=ALL_IMS,
```
joelridden, line 41: When using this, I'm having some issues with pSA, as it gets a KeyError with pSA in the columns. Maybe should add each im for pSA with the period to this?

jpa198: You want to explicitly list all 132 pSA periods?
Can I see the help output for that one? :P

joelridden (approved):

jpa198 (approved):

scripts/summarise_all_rels_im.py
```diff
@@ -73,60 +76,46 @@
...
-            modified_im_name = (
-                im_type if im_type[0] != "p" else f"p{im_type[1:].replace('p', '.')}"
-            )
-            df_dict[modified_im_name] = log_mean_im
-            df_dict[f"{modified_im_name}_sigma"] = log_stdev_im
+        merged_im_df = pd.concat(rel_im_dfs, axis=0, keys=range(len(rel_im_dfs)))
+        if stat_fn == "mean":
+            log_mean_im = np.exp(np.log(merged_im_df).mean(level=[1, 2]))
+        else:
+            log_mean_im = np.exp(np.log(merged_im_df).median(level=[1, 2]))
```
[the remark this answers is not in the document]
sungeunbae: good point

sungeunbae (approved):

jpa198 (approved):

PR #147, opened 2022-03-08
Vectorise GC2

joelridden:
Created a new function to do the vectorisation of GC2 across sites and hypocentre locations.
Just requires the extra parameter of origin offsets as a numpy array.

### DIFF
--- a/IM_calculation/source_site_dist/src_site_dist.py
+++ b/IM_calculation/source_site_dist/src_site_dist.py
@@ -189,7 +189,7 @@ def calc_rx_ry_GC2(
         # Will only use the first one found if there are multiple
         for plane in plane_infos:
             if plane["shyp"] != -999.9000:
-                origin_offset = length / 2 + plane["shyp"]
+                origin_offset = -(length / 2 + plane["shyp"])
                 break
 
     for i, loc in enumerate(locations):
--- a/IM_calculation/source_site_dist/src_site_dist.py
+++ b/IM_calculation/source_site_dist/src_site_dist.py
@@ -1,19 +1,20 @@
 from typing import List, Dict
 
 import matplotlib.path as mpltPath
-import numba
 import numpy as np
 
-from qcore.geo import get_distances, get_multiple_distances, ll_cross_along_track_dist, ll_bearing
+from qcore.geo import (
+    get_distances,
+    get_multiple_distances,
+    ll_cross_along_track_dist,
+    ll_bearing,
+)
 
-numba.config.THREADING_LAYER = "omp"
-h_dist_f = numba.njit(get_distances)
 
 VOLCANIC_FRONT_COORDS = [(175.508, -39.364), (177.199, -37.73)]
 VOLCANIC_FRONT_LINE = mpltPath.Path(VOLCANIC_FRONT_COORDS)
 
 
-@numba.jit(parallel=True)
 def calc_rrup_rjb(srf_points: np.ndarray, locations: np.ndarray):
     """Calculates rrup and rjb distance
 
@@ -36,15 +37,23 @@ def calc_rrup_rjb(srf_points: np.ndarray, locations: np.ndarray):
     rrups = np.empty(locations.shape[0])
     rjb = np.empty(locations.shape[0])
 
-    for loc_ix in numba.prange(locations.shape[0]):
-        h_dist = h_dist_f(srf_points, locations[loc_ix, 0], locations[loc_ix, 1])
+    # Size limit for memory capacity
+    size = 5000
+    split_locations = np.split(locations, np.arange(size, locations.shape[0], size))
 
-        v_dist = srf_points[:, 2] - locations[loc_ix, 2]
+    for ix, cur_locations in enumerate(split_locations):
+        h_dist = get_multiple_distances(
+            srf_points, cur_locations[:, 0], cur_locations[:, 1]
+        )
 
-        d = np.sqrt(h_dist ** 2 + v_dist ** 2)
+        v_dist = srf_points[:, 2, np.newaxis] - cur_locations[:, 2]
 
-        rrups[loc_ix] = np.min(d)
-        rjb[loc_ix] = np.min(h_dist)
+        d = np.sqrt(h_dist ** 2 + v_dist.T ** 2)
+
+        start_ix = size * ix
+        end_ix = start_ix + cur_locations.shape[0]
+        rrups[start_ix:end_ix] = np.min(d, axis=1)
+        rjb[start_ix:end_ix] = np.min(h_dist, axis=1)
 
     return rrups, rjb
 
@@ -115,7 +124,7 @@ def calc_rx_ry_GC1(
             # Get cumulative number of points in each plane
 
             # Get the closest point in the fault
-            h_dist = h_dist_f(srf_points, lon, lat)
+            h_dist = get_distances(srf_points, lon, lat)
 
             # Get the index of closest fault plane point
             point_ix = np.argmin(h_dist)
@@ -201,7 +210,7 @@ def calc_rx_ry_GC2(
             r_x_p, r_y_p = calc_rx_ry_GC1(
                 plane_points, [plane_header], np.asarray([loc])
             )
-            dists = h_dist_f(plane_points, loc[0], loc[1])
+            dists = get_distances(plane_points, loc[0], loc[1])
             # Mimimum distance of 0.001km to prevent nans/infs
             # A bit hacky but it works. Only needed when a location is directly on top of a subfault
             dists = np.maximum(dists, 0.001)
--- a/IM_calculation/source_site_dist/src_site_dist.py
+++ b/IM_calculation/source_site_dist/src_site_dist.py
@@ -3,13 +3,7 @@ from typing import List, Dict
 import matplotlib.path as mpltPath
 import numpy as np
 
-from qcore.geo import (
-    get_distances,
-    get_multiple_distances,
-    ll_cross_along_track_dist,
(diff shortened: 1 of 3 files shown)


IM_calculation/source_site_dist/src_site_dist.py
```diff
@@ -218,6 +218,66 @@ def calc_rx_ry_GC2(
     return r_x, r_y
 
 
+def calc_rx_ry_GC2_vectorised(
```
jpa198, line 221: "vectorised" is too generic as a suffix (The original function was already vectorised over sites :P). Suggest calling it multi_hypocentre or something, to show difference from the original algorithm

IM_calculation/source_site_dist/src_site_dist.py
```diff
@@ -218,6 +218,66 @@ def calc_rx_ry_GC2(
...
+        # Mimimum distance of 0.001km to prevent nans/infs
+        # A bit hacky but it works. Only needed when a location is directly on top of a subfault
+        dists = np.maximum(dists, 0.001)
+        weights_p = np.sum(np.power(dists, -2), axis=1)
+
+        weights += weights_p
+        r_x_values[:] += weights_p * r_x_p
+        # Reshape to allow broadcasting of the offsets to create the (offsets, locations) shape matrix
+        r_y_values[:] += weights_p * (
+            r_y_p.reshape(len(locations), 1).T + offsets.reshape(len(offsets), 1)
```
jpa198, line 271: How does this line work? From reading the documentation the first argument of the reshape method is the shape, and the second argument is the matrix order (C vs Fortran)

IM_calculation/source_site_dist/src_site_dist.py
jpa198, line 221: "vectorised" is too generic as a suffix (The original function was already vectorised over sites :P). Suggest calling it multi_hypocentre or something, to show difference from the original algorithm

IM_calculation/source_site_dist/src_site_dist.py
```diff
@@ -218,6 +218,66 @@ def calc_rx_ry_GC2(
...
+
+    offsets = origin_offsets
+    weights = np.zeros(len(locations))
+    r_x_values = np.zeros((len(offsets), len(locations)))
+    r_y_values = np.zeros((len(offsets), len(locations)))
+    for plane_points, plane_header in zip(pnt_sections, plane_infos):
+        r_x_p, r_y_p = calc_rx_ry_GC1(plane_points, [plane_header], locations)
+        dists = np.asarray(
+            [h_dist_f(plane_points, loc[0], loc[1]) for loc in locations]
+        )
```
jpa198, line 261: To remove the list comprehension here would require refactoring/duplicating h_dist_f to allow for m*1 by 1*n matrix operations. Off the top of my head it will currently try to treat inputs as vectors and only run pairwise operations

jpa198: There might be something you could do with newaxis to make it work though

IM_calculation/source_site_dist/src_site_dist.py
joelridden: So for an example with 2 sites 100 hypocentre locations when it reaches line 270:

weights_p = shape(2,)
r_y_p = shape(2,)
offsets = shape(100,)

To add each offset to r_y_p it would require the shape (1,2) and (100, 1).
Can't be done with (2,) and (100,) and so the reshape is required (I have removed the transpose and just swapped the len(locations) and 1 around as it was not needed)
image

Just to prove that reshape doesn't actually need the extra () the 1 it's doesn't go into the matrix order parameter.
image

This creates a (100, 2) array which is then multiplied by the weights of (2,) and that works normally.
Hopefully this should be enough to explain it. (I don't think the comment needs more information?)

jpa198: Huh, the magic of broadcasting. I'd swap the order of r_y_p and offsets to make it (100, 1) + (1, 2) = (100, 2). But that's just me

jasonmotha (dismissed):

IM_calculation/source_site_dist/src_site_dist.py
```diff
@@ -218,6 +218,64 @@ def calc_rx_ry_GC2(
...
 
 
+def calc_rx_ry_GC2_multi_hypocentre(
+    srf_points: np.ndarray,
+    plane_infos: List[Dict],
+    locations: np.ndarray,
+    origin_offsets: np.ndarray = np.asarray([0]),
+):
+    """
+    Vectorised version of the GC2 function along multiple hypocentre locations.
```
claudio525, line 228: Comment/Docstring should be (roughly) in line with code line length imo

IM_calculation/source_site_dist/src_site_dist.py
```diff
@@ -218,6 +218,64 @@ def calc_rx_ry_GC2(
...
+    according to plane length and distance to the location
+    For one fault plane this is the same as the GC1 function
+    :param srf_points: An array with shape (n, 3) giving the lon, lat, depth location of each subfault
+    :param plane_infos: A list of srf header dictionaries, as retrieved from qcore.srf.get_headers with idx=True
+    :param locations: An array with shape (m, 2) giving the lon, lat locations of each location to get Rx, Ry values for
+    :param origin_offsets: An array with shape (o) with the along strike hypocentre locations from the centre of
+    the fault for multiple realisations. If not set then only one origin will be considered at the centre position.
+    :return: Two arrays with shape (o, m) giving the Rx, Ry values for each of the given locations for each hypocentre
+    """
+    # Separate the srf points into the different plane traces
```
claudio525, line 240: Would suggest adding some more block comments, since I this is not a trivial function. Whoever has to touch this next will appreciate it. Up to you though.

IM_calculation/source_site_dist/src_site_dist.py
```diff
@@ -218,6 +218,64 @@ def calc_rx_ry_GC2(
     return r_x, r_y
 
 
+def calc_rx_ry_GC2_multi_hypocentre(
+    srf_points: np.ndarray,
+    plane_infos: List[Dict],
+    locations: np.ndarray,
+    origin_offsets: np.ndarray = np.asarray([0]),
+):
```
jpa198, line 226: I think the existing function can be turned into a wrapper which calls this in the case of only one hypocentre. I don't think it makes sense to have two copies of almost the same function when it can be reduced to one copy.

IM_calculation/source_site_dist/src_site_dist.py
```diff
@@ -1,19 +1,20 @@
 from typing import List, Dict
 
 import matplotlib.path as mpltPath
-import numba
 import numpy as np
 
-from qcore.geo import get_distances, get_multiple_distances, ll_cross_along_track_dist, ll_bearing
+from qcore.geo import (
```
jasonmotha, line 6: Can we change that to import the geo module?

jasonmotha (commented):
Looks good.

Can't see the specification for float32 though?

Can you do a quick memory test with a WairarapaNich or Hik*max SRF and 27k stations?

our 5k might be a litte underconservative.

jasonmotha (approved):

joelridden:
> Looks good.
> 
> Can't see the specification for float32 though?
> 
> Can you do a quick memory test with a WairarapaNich or Hik*max SRF and 27k stations?
> 
> our 5k might be a litte underconservative.

Ran for fault WairarapNich with 1k stations per loop and the function was using at max 16GB of ram

IM_calculation/source_site_dist/src_site_dist.py
```diff
@@ -1,19 +1,14 @@
...
-
-numba.config.THREADING_LAYER = "omp"
-h_dist_f = numba.njit(get_distances)
+from qcore import geo
 
 VOLCANIC_FRONT_COORDS = [(175.508, -39.364), (177.199, -37.73)]
 VOLCANIC_FRONT_LINE = mpltPath.Path(VOLCANIC_FRONT_COORDS)
 
 
-@numba.jit(parallel=True)
```
jpa198, line 16: :grinning:

IM_calculation/source_site_dist/src_site_dist.py
jpa198: Is there a plan to turn one of these functions into a wrapper for the other (either way)? I don't think there's a need for both to have explicit implementations.

IM_calculation/source_site_dist/src_site_dist.py
```diff
@@ -34,15 +28,18 @@ def calc_rrup_rjb(srf_points: np.ndarray, locations: np.ndarray):
...
-    rjb = np.empty(locations.shape[0])
+    rrups = np.empty(locations.shape[0], dtype=np.float32)
+    rjb = np.empty(locations.shape[0], dtype=np.float32)
+
+    srf_points = srf_points.astype(np.float32)
+    locations = locations.astype(np.float32)
 
     # Size limit for memory capacity
-    size = 5000
+    size = 1000
```
claudio525, line 38: I think this should either be defined as a constant (uppercase) right at the top of the function, or make it a parameter of the function. I prefer the latter approach. Also maybe a more descriptive name, such as `n_stations_per_iter`

jasonmotha: A function parameter makes a lot of sense for this. I didn't like it but thought a constant was a bit overkill.

jasonmotha (approved):

claudio525 (approved):

jpa198 (dismissed):

claudio525 (approved):

PR #148, opened 2022-03-31
Rejig run.py to work with the current Advanced IM framework

jpa198:
This can be transplanted into place over the previous run.py

This should be compared with main() in https://github.com/ucgmsim/IM_calculation/blob/master/IM_calculation/Advanced_IM/runlibs_2d.py as it's based on that, with accommodations for the way 3D models work compared with 2D models.
Could probably be substantially abstracted in case more 3D models are added

### DIFF
--- a/IM_calculation/Advanced_IM/Models/Spear_3D_Model/run.py
+++ b/IM_calculation/Advanced_IM/Models/Spear_3D_Model/run.py
@@ -50,6 +50,11 @@ def main(NS_component, EW_component, OpenSees_path, output_dir, im_name, run_scr
     # saves starting time
     datetime_to_file(datetime.now(), time_type.start_time.name, output_dir)
 
+
+    for comp in BASIC_HORIZONTAL_COMPONENTS:
+        # Give the sub components start times
+        shutil.copy(os.path.join(output_dir, time_type.start_time.name), os.path.join(output_dir, comp.str_value))
+
     try:
         subprocess.run(script, timeout=timeout_threshold)
     except subprocess.TimeoutExpired:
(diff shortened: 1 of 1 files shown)


jpa198:
Looks like tests are failing due to the src_site distance changes

sungeunbae:
> Looks like tests are failing due to the src_site distance changes

Do you mean we should update the test too?

jpa198:
> > Looks like tests are failing due to the src_site distance changes
> 
> Do you mean we should update the test too?

No, I think #147 just has to go in first and those will be sorted out

jpa198:
This is also blocked pending a field test with Vahid, I'll try and get it sorted today though

IM_calculation/Advanced_IM/Models/Spear_3D_Model/run.py
```diff
@@ -146,13 +197,13 @@ def calculate_geom(output_dir, im_name):
...
     """
-    df_000 = pd.read_csv(
-        os.path.join(output_dir, f"{im_name}_000.csv"), dtype={"component": str}
-    )
-    df_090 = pd.read_csv(
-        os.path.join(output_dir, f"{im_name}_090.csv"), dtype={"component": str}
-    )
-    #
+
+    c000_path = os.path.join(output_dir, f"{im_name}_000.csv")
```
sungeunbae, line 201: It appears that you chose to go with NS, EW because they are more self-explanatory than 000 and 090, but there are places in code still sticking to 000 and 090. Have you considered keeping things consistent?

jasonmotha (approved):
:rofl:

IM_calculation/Advanced_IM/Models/Spear_3D_Model/run.py
jpa198: Good point, I've changed it back

jasonmotha (approved):

sungeunbae (approved):

PR #149, opened 2022-04-19
Fix hypocentre offset math

jpa198:
Offset should be taken from the middle of the entire fault trace, not from the middle of the plane trace
The original logic seemed to be broken too


claudio525 (dismissed):

jasonmotha (approved):

claudio525 (approved):

jasonmotha (approved):
A+/10

claudio525 (approved):

PR #153, opened 2022-08-03
Sdof imk rotd

sungeunbae:
Upon @Vahid-Loghman 's request. (Tested ok)
Lots of tcl scripts (which were supplied by Vahid), but the most important part is the separation of run_and_check from the main function in runlibs_2d.py

The original main was designed for conventional 000, 090 components, computing each component at a time, while rotd-based IMs don't need the component-wise calculation. The main() also does extra bits like computing geom component, and producing csv output etc., which is not relevant to this IM.

The run.py of this new Adv IM calls this new function run_and_check(), not the main(), and produces output file the way it wants. Perhaps we could organize a separate main function for rotd-based IMs, but I don't think we have the critical mass yet.


jpa198:
First big question, why is this being run as a separate IM to the orthogonal components? I would have expected this to be a good opportunity to add rotd handling to adv_im. (Even if it didn't get used much)

IM_calculation/Advanced_IM/Models/SDOF_IMK_RotD/run.py
```diff
@@ -0,0 +1,63 @@
...
+
+    model_converged = runlibs_2d.run_and_check(run_script,
+                                               [getattr(args, component_list[0].str_value),getattr(args, component_list[1].str_value)],
+                                               str(output_dir),args.OpenSees_path, args.timeout_threshold)
+
+    assert model_converged, f"{output_dir} failed to converge."
+
+    with open(os.path.join(output_dir, "max_rotD_resp.txt"), "r") as f:
+        data = f.readlines()
+    rotd_values = im_calculation.get_rotd_components_dict([float(x) for x in data[0].split(" ")], rotd_comps)
```
jasonmotha, line 43: Would be good to have a comment that you explicitly want to split on `" "` rather than any whitespace with `.split()`

IM_calculation/Advanced_IM/runlibs_2d.py
```diff
@@ -89,15 +90,50 @@ def datetime_to_file(t_value, t_type: str, out_dir):
 
 # def save_timeout_value(component_outdir, timeout_threshold, datetime.datetime.now()):
 
+def run_and_check(run_script, comp_paths, output_dir, opensees_path, timeout):
```
jpa198, line 93: Have the existing models been tested with this change? I wouldn't expect any issues, but a concrete check would be good

IM_calculation/Advanced_IM/advanced_im_config.yaml
```diff
@@ -46,4 +46,6 @@ Haselton_RC_20Story:
 SDOF_IMK:
   script_location: 'Models/SDOF_IMK/run.py'
   timeout_threshold: null
-  
\ No newline at end of file
+SDOF_IMK_RotD:
```
jasonmotha, line 49: Storing a component in the IM name seems like a different style to how the rest of the IMs work. I'll need a deeper dive into this code - but I'm curious to know why you chose this architecture.

What's the difference in the TCL between SDOF_IMK and SDOF_IMK_RotD?

jpa198: The TCL itself handles the rotd logic (projecting the timeseries into the rotd directions) instead of doing it via python. Hence my question as to why we're doing it via TCL instead of via python like for simple IMs.

IM_calculation/Advanced_IM/Models/SDOF_IMK_RotD/run.py
sungeunbae: > Would be good to have a comment that you explicitly want to split on `" "` rather than any whitespace with `.split()`

Good point. I think split() will work just fine. Will update it in the next commit.

IM_calculation/IM/im_calculation.py
```diff
@@ -94,6 +94,19 @@ def check_rotd(comps_to_store: Iterable[Components]) -> bool:
...
+    rotd50 = np.median(rotd, axis=-1)
+    rotd100 = np.max(rotd, axis=-1)
+
+    if Components.crotd50 in comps_to_store:
+        value_dict[Components.crotd50.str_value] = rotd50
+    if Components.crotd100 in comps_to_store:
+        value_dict[Components.crotd100.str_value] = rotd100
+    if Components.crotd100_50 in comps_to_store:
+        value_dict[Components.crotd100_50.str_value] = rotd100 / rotd50
+    return value_dict
```
jpa198, line 109: This is out again

sungeunbae: Yes, it is intentional. This seperate function was needed for Vahid's calculation (which is urgent), which uses Tcl code for getting rotation. I just needed a way to return the rotd components. This branch is not going to be merged until I make it not depend on Tcl for rotation, which will make this separation unnecessary.

jpa198: Is this change still in progress?

PR #155, opened 2022-10-28
MPI IM Calc

joelridden:
Changing from Multiprocessing to MPI
Computation briefly explained:
1 master sending a single station to each other process.
Each process computes and writes to a single station file
After all stations are computed then the master compiles all the station files together

Can be checkpointed per station if stopped halfway as it reads the output directory and removes stations that are already done.
Both Advanced IM and standard run were tested on Maui and outputs were checked.

Relies on Qcore and Workflow PR to go through to function correctly.
https://github.com/ucgmsim/qcore/pull/275
https://github.com/ucgmsim/slurm_gm_workflow/pull/456

### DIFF
--- a/IM_calculation/IM/im_calculation.py
+++ b/IM_calculation/IM/im_calculation.py
@@ -673,7 +673,7 @@ def compute_measures_mpi(
             lambda: comm.sendrecv(None, dest=master), StopIteration
         ):
             logger.info(f"Station to compute: {station}")
-            waveform = read_waveform.read_waveforms(
+            waveforms = read_waveform.read_waveforms(
                 input_path,
                 bbseries,
                 [station],
@@ -685,10 +685,10 @@ def compute_measures_mpi(
             logger.info(f"Finished reading waveform: {station}")
             # only run basic im if and only if adv_im not going to run
             if running_adv_im:
-                compute_adv_measure(waveform, advanced_im_config, output)
+                compute_adv_measure(waveforms, advanced_im_config, output)
             else:
                 result_dict = compute_measure_single(
-                    waveform,
+                    waveforms[0],
                     sorted(ims),
                     sorted(components_to_store, key=lambda x: x.value),
                     im_options,
--- a/IM_calculation/IM/im_calculation.py
+++ b/IM_calculation/IM/im_calculation.py
@@ -560,9 +560,12 @@ def compute_measures_mpi(
                 file_type=file_type,
                 units=units,
             )
+            print("Finished waveform readings")
             # only run basic im if and only if adv_im not going to run
             if running_adv_im:
+                print("Running Adv Measures")
                 compute_adv_measure(waveforms, advanced_im_config, output)
+                print("Finished Adv Measures")
             else:
                 result_dict = compute_measure_single(
                     waveforms[0],
@@ -575,6 +578,7 @@ def compute_measures_mpi(
                 )
                 write_result(result_dict, station_path, station, simple_output)
     if is_master:
+        print("Finished MPI computation")
         if running_adv_im:
             # read, agg and store csv
             advanced_IM_factory.agg_csv(advanced_im_config, station_names, output)
@@ -584,7 +588,9 @@ def compute_measures_mpi(
                 get_result_filepath(output, identifier, ".csv"), index=False
             )
             shutil.rmtree(station_path)
+        print("Agged csv")
         generate_metadata(output, identifier, rupture, run_type, version)
+        print("generated Metadata")
 
 
 def get_result_filepath(output_folder, arg_identifier, suffix):
--- a/IM_calculation/IM/im_calculation.py
+++ b/IM_calculation/IM/im_calculation.py
@@ -142,9 +142,7 @@ def compute_adv_measure(waveform, advanced_im_config, output_dir):
     :param output_dir: Directory where output folders are contained. Structure is /path/to/output_dir/station/im_name
     :return:
     """
-    print(waveform)
     waveform_acc = waveform[0]
-    print(waveform_acc)
     station_name = waveform_acc.station_name
     adv_im_out_dir = os.path.join(output_dir, station_name)
     advanced_IM_factory.compute_ims(waveform_acc, advanced_im_config, adv_im_out_dir)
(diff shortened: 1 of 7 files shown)


Jenkinsfile
```diff
@@ -53,6 +53,13 @@ pipeline {
                     ln -s $HOME/data/testing/${env.JOB_NAME}/sample0/output sample0
                     echo "[ Run test now ]"
                     pytest -s
+                    echo "[ Linking test data ]"
+                    rm -rf sample0
+                    mkdir sample0
+                    ln -s $HOME/data/testing/${env.JOB_NAME}/sample0/input sample0
+                    ln -s $HOME/data/testing/${env.JOB_NAME}/sample0/output sample0
```
jpa198, line 60: Why do you nuke the test data for the MPI test?

joelridden: The test data gets nuked after running pytest -s
Then need to recreate sample0 and relink again to run the mpi test

sungeunbae:
Does it always need to be run with mpirun?

joelridden:
> Does it always need to be run with mpirun?

Yes, have tried without but it was needed. Documentation here https://pytest-mpi.readthedocs.io/en/latest/usage.html

Jenkinsfile
jpa198: :( If we aren't downloading the data in python, we shouldn't be deleting it either... 
This is fine for now though

IM_calculation/IM/im_calculation.py
```diff
@@ -507,63 +511,76 @@ def compute_measures_multiprocess(
...
+            if len(stations_to_run) > 0:
+                station = stations_to_run.pop(-1)
+                logger.info(f"Sending station {station} to {slave_id}")
+                comm.send(obj=station, dest=slave_id)
+            else:
+                comm.send(obj=StopIteration, dest=slave_id)
+                nslaves -= 1
+        logger.info("All stations complete")
+    else:
+        for station in iter(lambda: comm.sendrecv(None, dest=master), StopIteration):
```
jpa198, line 547: The last command sent is stop iteration, does it need to be in the worker for loop too?

IM_calculation/IM/im_calculation.py
```diff
@@ -507,63 +511,76 @@ def compute_measures_multiprocess(
...
+    station_path = Path(output) / "stations"
+    stations_set = set(station_names)
+    found_stations = {
+        file.stem for file in station_path.iterdir() if file.stat().st_size > 0
+    }
+    stations_to_run = list(stations_set.difference(found_stations))
+
+    status = MPI.Status()
+    if is_master:
+        nslaves = size - 1
```
jpa198, line 533: Should be using server/worker instead of master/slave

jpa198, line 533: Do we ever check that size is greater than 1? Otherwise there is only the server and no workers, so nothing actually gets done

IM_calculation/scripts/calculate_ims.py
```diff
@@ -10,12 +10,27 @@
...
+    from qcore import MPIFileHandler
+
+    comm = MPI.COMM_WORLD
+    rank = comm.Get_rank()
+    size = comm.Get_size()
+    master = 0
+    is_master = not rank
+
+    logger = logging.getLogger("IM_calc_rank_%i" % comm.rank)
+    logger.setLevel(logging.DEBUG)
```
jpa198, line 33: I don't think calculate_ims gets imported anywhere, so this shouldn't need a name main wrapper
Doesn't hurt though

IM_calculation/scripts/calculate_ims.py
```diff
@@ -135,6 +150,14 @@ def load_args():
...
     )
 
+    parser.add_argument(
+        "-nd",
+        "--nodes",
+        default=1,
+        type=int,
+        help="Please provide the number of processors. Default is 1",
+    )
+
```
jpa198, line 160: Wut? You're in MPI land

IM_calculation/scripts/calculate_ims.py
```diff
@@ -219,9 +251,11 @@ def main():
         im_options["FAS"] = calc.validate_fas_frequency(args.fas_frequency)
 
     # Create output dir
-    utils.setup_dir(args.output_path)
-    if not args.simple_output and args.advanced_ims is None:
-        utils.setup_dir(os.path.join(args.output_path, calc.OUTPUT_SUBFOLDER))
+    if is_master:
+        utils.setup_dir(args.output_path)
+        utils.setup_dir(Path(args.output_path) / "stations")
```
jpa198, line 256: If you set the type of "--output_path" to Path, you don't need to do this down here

jpa198, line 256: Only the second of the two needs to run, it'll make the first one as a side effect

IM_calculation/IM/im_calculation.py
```diff
@@ -454,9 +450,12 @@ def get_bbseis(input_path, file_type, selected_stations, real_only=False):
     return bbseries, station_names
 
 
-def compute_measures_multiprocess(
+def compute_measures_mpi(
     input_path,
     file_type,
+    comm,
+    rank,
+    size,
```
jpa198, line 458: rank and size can be derived from comm? Not much point in passing them through

IM_calculation/IM/im_calculation.py
```diff
@@ -2,19 +2,19 @@
...
 import sys
+import shutil
 from datetime import datetime
 from functools import partial
-from multiprocessing.pool import Pool
-from collections import ChainMap
+from pathlib import Path
 from typing import List, Iterable
 
+from mpi4py import MPI
```
jpa198, line 11: no name main wrapper for this? Makes it impossible to import any of the individual IM functions without MPI running in the background

jpa198 (commented):
Lot of comments, not many of them are actual issues though

jpa198:
FYI we're going to look at this again on Monday, we didn't really think about the desktop use case for running IM_calc. Might have to have the MP version in parallel or something

sungeunbae (dismissed):

jpa198 (approved):

sungeunbae (approved):

PR #160, opened 2023-06-07
Added simple IM calculation example, plus some minor edits

claudio525:


sungeunbae (approved):

joelridden (approved):
Do we want / need an example from a BB file? Can't really store the BB file on git though. Is that the more common use case or are these waveform files per component that can be read with pandas more common?

claudio525:
@joelridden I think this one is fine for now, as its nice and general, i.e. if they have another data format they just need to change the data loading part.

PR #161, opened 2023-06-15
SNR calculation + KO Structure

joelridden:
Can calculate FAS, SNR for a waveform.
Adjustments to KO matrices creation to remove duplicated code and ensure correct placement of matrices files.

Added example script to calculate SNR for waveform data, also plots the output (shown below)
image

Verification of a similar curve was done in comparison to a version in the GM_classifier (Some minor differences, tapering etc)


IM_calculation/IM/snr_calculation.py
```diff
@@ -0,0 +1,107 @@
+import numpy as np
+from scipy.interpolate import interp1d
+
+import IM_calculation.IM.computeFAS as computeFAS
+from IM_calculation.IM.read_waveform import Waveform
+
+
+def apply_taper(acc, percent=0.05):
```
claudio525, line 8: type hints (since you are using them below)

IM_calculation/scripts/A_KonnoMatricesComputation.py
```diff
@@ -1,47 +1,74 @@
...
+    :param num_to_gen: Number of matrices to generate
+    :param bandwidth: Bandwidth of Konno Ohmachi smoothing
+    """
+    Path(install_directory).mkdir(exist_ok=True)
     for i in range(num_to_gen):
         # n = [512, 1024, 2048, 4096, 8192, 16384, 32768]
         n = 512 * 2 ** i
-        file_name = os.path.join(root, f"KO_{n}.npy")
-        matrix = createKonnoMatrix_single(n * 2)
+        file_name = install_directory / f"KO_{n}.npy"
```
claudio525, line 34: If install directory is a string this will crash I think?

claudio525 (commented):
Looks good, just a couple of minor comments

IM_calculation/scripts/A_KonnoMatricesComputation.py
joelridden: Turns out it never is called with a string anymore so safe to make the typehint a Path instead

claudio525 (approved):

sungeunbae (approved):

PR #162, opened 2023-06-22
KO directory update and FAS Interpolate

joelridden:
Allows for a custom location of KO matrices as when getting up to large factors 9 etc you get to file sizes of 171 GB's.
Does interpolation of FAS as well as SNR if a common frequency vector is specified.


sungeunbae (approved):
Looking good. Somewhere in the code, can you leave a little warning in the code saying things can go wild?

claudio525 (approved):

PR #165, opened 2024-03-15
Added cutoff for SNR / FAS

joelridden:
Allows for the ability to turn on and off smoothing from KO matrices
Applies the cutoff to nan after values going beyond the sampling rate of the data to avoid negative values

Based on the notes here:
[Note on FAS Computation.pdf](https://github.com/ucgmsim/IM_calculation/files/14657688/Note.on.FAS.Computation.pdf)

### DIFF
--- a/IM_calculation/IM/im_calculation.py
+++ b/IM_calculation/IM/im_calculation.py
@@ -151,7 +151,7 @@ def compute_measure_single(
     comps_to_store,
     im_options,
     comps_to_calculate,
-    progress = None,
+    progress=None,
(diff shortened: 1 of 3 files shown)


sungeunbae:
Do you want to assign reviewers?

joelridden:
> Do you want to assign reviewers?

Not right now as there is some test issues and need to clean up some testing I added when I was running into errors on hypocentre. I thought this was complete but not quite yet. Have added it to a draft for now

sungeunbae (approved):

claudio525 (approved):

PR #167, opened 2024-05-10
Handle new txt func

joelridden:
Better docstrings and handles new timeseries function dealing with the filename

### DIFF
--- a/IM_calculation/Advanced_IM/advanced_IM_factory.py
+++ b/IM_calculation/Advanced_IM/advanced_IM_factory.py
@@ -1,6 +1,6 @@
+from pathlib import Path
 from collections import namedtuple
 
-import numpy as np
 import pandas as pd
 import os
 import re
@@ -107,7 +107,7 @@ def save_waveform_to_tmp_files(
         timeseries_to_text(
             accelerations.values[:, COMP_DICT[component]],
             accelerations.DT,
-            filename,
+            Path(filename),
             station_name,
             component,
         )
--- a/IM_calculation/Advanced_IM/advanced_IM_factory.py
+++ b/IM_calculation/Advanced_IM/advanced_IM_factory.py
@@ -22,7 +22,18 @@ COMP_DICT = {"090": 0, "000": 1, "ver": 2}
 
 
 def get_config(config_file=CONFIG_FILE_NAME):
-    return yaml.safe_load(open(config_file))
+    """
+    Loads the configuration file and returns the dictionary
+
+    Parameters
+    ----------
+    config_file : str, optional
+        Path to the configuration file
+        default: advanced_im_config.yaml
+    """
+    with open(config_file) as yaml_file:
+        loaded_config = yaml.safe_load(yaml_file)
+    return loaded_config
 
 
 def get_im_list(config_file=CONFIG_FILE_NAME):
--- a/IM_calculation/Advanced_IM/advanced_IM_factory.py
+++ b/IM_calculation/Advanced_IM/advanced_IM_factory.py
@@ -2,13 +2,13 @@ from pathlib import Path
 from collections import namedtuple
 
 import pandas as pd
+import yaml
 import os
 import re
 import subprocess
 import tempfile
 
 from qcore import timeseries
-from qcore import utils
 from IM_calculation.IM.read_waveform import Waveform
 
 
@@ -22,7 +22,7 @@ COMP_DICT = {"090": 0, "000": 1, "ver": 2}
 
 
 def get_config(config_file=CONFIG_FILE_NAME):
-    return utils.load_yaml(config_file)
+    return yaml.safe_load(open(config_file))
 
 
 def get_im_list(config_file=CONFIG_FILE_NAME):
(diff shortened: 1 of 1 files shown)


claudio525 (approved):

IM_calculation/Advanced_IM/advanced_IM_factory.py
```diff
@@ -67,23 +71,43 @@ def compute_ims(accelerations, configuration, adv_im_out_dir):
...
 
-def get_acc_filename(folder, stat, component):
-    # returns the same filename structure as timeseries.seis2txt
+def get_acc_filename(folder: str, stat: str, component: str):
+    """
+    Returns the filename for the acceleration file
+
+    Parameters
+    ----------
+    folder : str
```
sungeunbae, line 80: Could this be Path? I believe the rest of code is not really pathlib ready?

joelridden: I mainly didnt change the input type because this function is still used elsewhere in the code. My concern is my task of fixing the filename change to the qcore repo then expands into changing more things down the rabbithole as we go further and further

IM_calculation/Advanced_IM/advanced_IM_factory.py
```diff
@@ -1,12 +1,16 @@
...
 from collections import namedtuple
+
 import pandas as pd
 import os
 import re
 import subprocess
 import tempfile
 
-from qcore.timeseries import seis2txt
+from qcore.timeseries import timeseries_to_text
```
lispandfound, line 10: This should just import the timeseries module not the function within the module

IM_calculation/Advanced_IM/advanced_IM_factory.py
```diff
@@ -1,12 +1,16 @@
...
 from qcore.utils import load_yaml
```
lispandfound, line 11: Not a you problem, but this `load_yaml` function is completely unnecessary now that Python has standardised the behaviour that `dict` keys are always ordered (indeed, `OrderedDict` has a completely different purpose now that `load_yaml` is almost certainly not considering properly). You can without loss of generality replace all instance of `load_yaml` with `yaml.safe_load`.

lispandfound (changes requested):

IM_calculation/Advanced_IM/advanced_IM_factory.py
joelridden: We could make this an issue for the repository, more of a code clean up task than actually fixing anything unless it hindered performance for example which in this case I don't think so, so I'd say this is pretty low priority.

claudio525 (dismissed):

IM_calculation/Advanced_IM/advanced_IM_factory.py
```diff
@@ -18,7 +22,7 @@
 
 
 def get_config(config_file=CONFIG_FILE_NAME):
-    return load_yaml(config_file)
+    return utils.load_yaml(config_file)
```
lispandfound, line 25: Continuing my crusade against `qcore.utils.load_yaml`. It's redundant, just replace it with `yaml.safe_load`

IM_calculation/Advanced_IM/advanced_IM_factory.py
lispandfound: Do we even need this function at all??? Surely in the one place it's being used we can just do `folder / f'{stat}.{component}'`

lispandfound: but yeah I agree, we can just keep it how it is, if we fixed the whole repo every time we would never get anything done.

IM_calculation/Advanced_IM/advanced_IM_factory.py
```diff
@@ -67,23 +71,43 @@ def compute_ims(accelerations, configuration, adv_im_out_dir):
...
+    tmp_folder : str
+        Folder to save the files
+    accelerations : Waveform
+        Acceleration waveform with all components
+    station_name : str
+        Station name
     """
     for component in COMP_DICT.keys():
-        seis2txt(
+        filename = get_acc_filename(tmp_folder, station_name, component)
```
lispandfound, line 106: Buuuutttt, I would get rid of this reference to `get_acc_filename`

lispandfound (changes requested):

sungeunbae (approved):

IM_calculation/Advanced_IM/advanced_IM_factory.py
```diff
@@ -22,7 +22,7 @@
 
 
 def get_config(config_file=CONFIG_FILE_NAME):
-    return utils.load_yaml(config_file)
+    return yaml.safe_load(open(config_file))
```
lispandfound, line 25: This should probably be encapsulated in a with statement right?

lispandfound (changes requested):

IM_calculation/Advanced_IM/advanced_IM_factory.py
joelridden: tmp_folder is now a Path

lispandfound (approved):

sungeunbae (approved):

claudio525 (approved):

PR #169, opened 2024-07-10
Install fix

joelridden:
This removes the errors we have sometimes get such as numpy is not installed when trying to build the KO matrices.
Solution was to just remove the need for KO matrices to be generated on setup. If the user wants to do FAS then they will need to run the script themselves which is not a big deal. Means this will take up less space too as the KO matrices get quite large at larger numbers. Should make IM Calc easier for people to install and use.

Also adds pyproject to solve the cython and numpy issues.

https://uceqeng.slack.com/archives/C3SGQHXT8/p1698977035565129


claudio525 (approved):

sungeunbae (commented):
Happy with the change, we have konno_setup,py anyway. Is it possible to give a warning message to the user to advise to run konno_setup.py if the user tries to compute FAS when konno matrices are missing

AndrewRidden-Harper (approved):

lispandfound (approved):
Very happy to have a `pip install` friendly IM Calc! Will make integrating into the workflow much easier.

claudio525 (approved):

sungeunbae (approved):

PR #171, opened 2024-08-05
Compute 256 KO

joelridden:
fmin investigation required a KO_256.npy file which currently would not be computed by this script regardless of the number set to generate.
To solve this issue I reduced the starting value and then increased the num_to_gen variable to compensate.
Also changed docstring style while I was here.

### DIFF
--- a/IM_calculation/scripts/A_KonnoMatricesComputation.py
+++ b/IM_calculation/scripts/A_KonnoMatricesComputation.py
@@ -8,8 +8,13 @@ from obspy.signal.konnoohmachismoothing import calculate_smoothing_matrix
 def createKonnoMatrix_single(ft_len: int, bandwidth: int = 20):
     """
     Creates a single Konno Ohmachi matrix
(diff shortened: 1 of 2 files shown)


AndrewRidden-Harper (approved):
looks good

claudio525 (approved):

sungeunbae (approved):

sungeunbae:
At first, I felt slightly uncomfortable with the default value 8, as it could consume quite a large space especially /nesi/project - but only few people will need to have konno matrices generated anyway (thanks to the recent change).

sungeunbae:
> > At first, I felt slightly uncomfortable with the default value 8, as it could consume quite a large space especially /nesi/project - but only few people will need to have konno matrices generated anyway (thanks to the recent change) and less chance of /nesi/project running out of space.
> 
> The file size impact for this PR is only 516KB as we are generating one of the very small matrices

Oh, excellent. No worries then!

lispandfound (approved):

PR #173, opened 2024-09-30
SNR Overuse Core Issue + KO Matrix Path

joelridden:
This fixes the issue when running snr_calculation and probably im_calculation when computing FAS under multiprocessing it trys to utilize every core per process and so sends the load average into ridiculous numbers above 400 on hypocentre that I was seeing.
However I'm not 100% familiar with all the details of using threadpool_limits(limits=1, user_api="blas") here. Keen for thoughts on this, can test other methods, but this is the first method I found that works without major re-works.

Added the ability to specify your own ko_matrix_path for IM calculations which is useful for the NZGMDB and probbaly general use when you dont want your large KO files in your IM_calc environment location due to space.
I did not clean up all of the docstrings for this where there was no docstring, this entire repo needs a good clean but it's not top priority at the moment.

Added extra ignore for snr calc that was missed.

### DIFF
--- a/IM_calculation/IM/computeFAS.py
+++ b/IM_calculation/IM/computeFAS.py
@@ -16,9 +16,8 @@ from threadpoolctl import threadpool_limits
 
 matrices = {}
 matrix_lock = Lock()
-
-# Limit the number of threads used by numpy and scipy
-threadpool_limits(limits=1, user_api='blas')
+# Limit the number of threads for loading KO matrices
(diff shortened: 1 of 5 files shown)


sungeunbae (approved):
Interesting fix!

AndrewRidden-Harper (approved):

lispandfound (approved):

claudio525 (approved):

PR #179, opened 2025-01-27
Add intensity measure units and utilities

lispandfound:
This PR extracts the most salient points from https://github.com/ucgmsim/workflow/pull/14 to allow the user to write intensity measure calculation datasets to HDF5 files with descriptions and units attached. It also supports reading the intensity measures with units, so that researchers do not have to learn pint or friends to take advantage of the unit-aware calculations of the new format.


IM/im_reader.py
```diff
@@ -1,75 +1,93 @@
-"""Utility module for reading intensity measure files stored in HDF5 format."""
-
 from pathlib import Path
 
-import pandas as pd
+import pint_xarray  # noqa: F401
```
joelridden, line 3: Why should we ignore this for deptry?

tests/test_im_reader.py
```diff
@@ -1,48 +1,109 @@
 from pathlib import Path
 
-import pandas as pd
+import numpy as np
 import pytest
 import xarray as xr
 
-from IM.im_reader import IMFile
-from IM.ims import IM
+from IM.im_reader import (  # Replace with actual module name
```
joelridden, line 7: Is the # Replace with actual module name
a note to self?

tests/test_im_reader.py
```diff
@@ -1,48 +1,109 @@
...
+    coords = {
+        "station": ["A", "B", "C"],
+        "component": ["000", "090", "ver", "geom", "rotd0", "rotd50", "rotd100"],
+        "period": np.linspace(0.01, 1.0, 10),
+    }
+    variables = {
+        "pSA": (["component", "station", "period"], data),
+        "PGA": (["component", "station"], np.random.rand(7, 3)),
+        "PGV": (["component", "station"], np.random.rand(7, 3)),
+        "CAV": (["component", "station"], np.random.rand(7, 3)),
```
joelridden, line 28: Do we want to be testing CAV5 too?

tests/test_im_reader.py
lispandfound: Yeah I wasn't sure I was going to keep im_reader as the module for this so I used a stand-in initially.

joelridden (dismissed):

AndrewRidden-Harper (approved):
very nice

IM/im_reader.py
```diff
@@ -1,75 +1,97 @@
...
+    IM.Ds575: "Significant duration (5-75%)",
+    IM.Ds595: "Significant duration (5-95%)",
+    IM.pSA: "Pseudo-spectral acceleration",
+    IM.FAS: "Fourier amplitude spectrum",
+}
 
-    Attributes
+
+IM_UNITS = {
+    IM.PGA: "g0",
```
sungeunbae, line 61: Add a comment about "g0" and its namespace collision with g.

joelridden (approved):

AndrewRidden-Harper (approved):

PR #181, opened 2025-04-07
Faster IM Calc for Obs data

joelridden:
Speed up improvements by allowing the use of no numexpr.
Deals with pykooh multiprocessing performance issues by switching back to old method of producing KO matrix files.
Improve FAS calculation speed under different scenarios of cores / number of stations to compute for.

Test on all IM's for a single waveform
Before time taken: 21.5s
After time taken: 6.79 s


IM/scripts/gen_ko_matrix.py
```diff
@@ -0,0 +1,41 @@
...
+        int, typer.Option(help="Number of Ko matrix generations to compute, starts at 8, 16, 32, 64, etc. Default of 14 (going upto KO matrix of 65536)",)
+    ] = 14,
+    bandwidth: Annotated[
+        int, typer.Option(help="Bandwidth of the Konno-Ohmachi smoothing window",)
+    ] = 40,
+    dt: Annotated[
+        float, typer.Option(help="Sampling interval of the time series",)
+    ] = 0.005,
+):
+    """Generate the Konno matrices for different window sizes."""
```
lispandfound, line 32:
```suggestion
@cli.from_docstring(app)
def main(
    output_dir: Annotated[
        Path,
        typer.Argument(
            exists=False,
            file_okay=False,
        ),
    ],
    num_to_gen: Annotated[
        int, typer.Option()
    ] = 14,
    bandwidth: Annotated[
        int, typer.Option()
    ] = 40,
    dt: Annotated[
        float, typer.Option()
    ] = 0.005,
):
    """Generate the Konno matrices for different window sizes.
    
    Parameters
    ---------------
    output_dir : Path
        Directory where the Konno matrices will be saved.
    num_to_gen : int
        Number of KO matrix generations to compute. The largest matrix will have `4 * 2 ** num_to_gen` rows and columns.
    bandwith : int
        Bandwith of the Konno-Ohmachi smoothing window.
    dt : float
        Sampling interval of the time series."""
```

IM/ko_matrices.py
```diff
@@ -0,0 +1,44 @@
...
+    Returns
+    -------
+    np.ndarray
+        The loaded Konno matrix.
+
+    Raises
+    ------
+    FileNotFoundError
+        If the required matrix file does not exist.
+    """
```
lispandfound, line 26:
```suggestion
def get_konno_matrix(size: int, directory: Path) -> np.memmap:
    """
    Retrieves the precomputed Konno matrix from a file.

    Parameters
    ----------
    size : int
        The size of the matrix to load.
    directory : Path
        Directory containing precomputed KO matrices.

    Returns
    -------
    np.memmap
        The loaded Konno matrix.

    Raises
    ------
    FileNotFoundError
        If the required matrix file does not exist.
    """
```

Just to emphasize that the array is memmap'd and not loaded into memory. There are also some subtle gotcha's with memmap vs regular numpy arrays and being aware of that in the type and docstring is useful.

IM/im_calculation.py
```diff
@@ -174,26 +180,30 @@ def calculate_ims(
             "NUMEXPR_NUM_THREADS",
             "NUMBA_MAX_THREADS",
             "NUMBA_NUM_THREADS",
-            "OPENBLAS_NUM_THREADS",
```
lispandfound, line 177: Why was this removed? If it isn't there, all numpy matrix multiplications will be multi-threaded by default.

lispandfound (changes requested):
How are we handling the differing `dt` values? To my understanding of this setup, we are generating KO matrices with `dt = 0.005` by default, are you just resampling the waveforms at that `dt`?

IM/im_calculation.py
joelridden: Missed this, would have been during some of my testing

joelridden:
> How are we handling the differing `dt` values? To my understanding of this setup, we are generating KO matrices with `dt = 0.005` by default, are you just resampling the waveforms at that `dt`?

As a note, dt does not matter for us when generating the KO matrix using numpy rfft

lispandfound (approved):

AndrewRidden-Harper (approved):

PR #182, opened 2025-05-02
SNR Taper Adjustment

joelridden:
Simpler approach to solving an issue with double tapers on SNR and the impacts of tapering at tp.
Now we adjust the signal window to include some of the noise, but only parts that actually will be dampened out by the taper of 5%.
Was going to add tests to ensure that we don't taper the value at tp, but would require returning the snr_signal values which is too much extra work, so instead added an assert which will do the same.


AndrewRidden-Harper (approved):

IM/snr_calculation.py
```diff
@@ -80,27 +77,32 @@ def calculate_snr(
     ValueError
         If the noise duration is less than 1s and so SNR can't be computed.
     """
+    # This extra time is to ensure that when a taper is applied, the signal part of the waveform
+    # is not affected by the tapering. The tapering is applied to the signal and noise separately.
+    tp_extra = 20 * waveform[:, tp:].shape[1] / 19 - waveform[:, tp:].shape[1]
```
lispandfound, line 82: This is a bit clearer IMO if you factor and then convert to decmals

```
tp_extra = 20 * nt / 19 - nt
              = (20 / 19 - 1) * nt
              = nt / 19
              or
              ~ 0.0526 * nt
```
Did you maybe mean `0.05 * nt`?
```suggestion
    tp_extra = waveform[:, tp:].shape[1] / 19
```

lispandfound (changes requested):

lispandfound (approved):

PR #184, opened 2025-05-29
Fixes pSA calcs using too much memory

AndrewRidden-Harper:
Cesar encountered an issue where the pSA calculation was being killed for using too much memory. 

The issue was that `comp_0 = newmark_estimate_psa(
        waveforms[:, :, Component.COMP_0.value],
        t,
        dt,
        w,
    )`

    
   (and the 90 and vertical components) were trying to process waveforms for all stations at once. `rotd_psa_values()` was already processing the stations in smaller batches that didn't overwhelm the memory, so I called `newmark_estimate_psa()` in the same way.
   
The outputs of some functions changed, so I also updated those tests. The test for the correct numerical value in `test_rotd_psa_values()` is not very robust, so suggestions are welcome

### DIFF
--- a/IM/ims.py
+++ b/IM/ims.py
@@ -2,10 +2,8 @@
 
 import gc
 import multiprocessing
-import sys
 import warnings
 from collections.abc import Callable
-from dataclasses import dataclass
 from enum import IntEnum, StrEnum
 from pathlib import Path
 from typing import Optional
@@ -46,19 +44,6 @@ class IM(StrEnum):
     FAS = "FAS"
 
 
-@dataclass
-class RotdPsaValuesOutput:
-    """Output data structure for the rotd_psa_values() function."""
-
-    rotd_psa: npt.NDArray[np.float32]
-    """ndarray of float32 with shape `(n_stations, n_periods, 3)`
-       Array containing minimum (rotd0), median (rotd50) and maximum (rotd100) PSA values."""
-    psa_comp_0_maxes: npt.NDArray[np.float32]
-    """Maximum absolute displacement values for the 000 component."""
-    psa_comp_90_maxes: npt.NDArray[np.float32]
-    """Maximum absolute displacement values for the 090 component."""
-
-
 @numba.njit(cache=True)
 def newmark_estimate_psa(
     waveforms: npt.NDArray[np.float32],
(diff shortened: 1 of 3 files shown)


IM/ims.py
```diff
@@ -245,27 +292,16 @@ def pseudo_spectral_acceleration(
...
-        dt,
-        w,
-    )
     # Step size is the minimum of either the CPU count, or the maximum number of
     # steps that fits within the psa_rotd_maximum_memory_allocation.
     if psa_rotd_maximum_memory_allocation:
         step = min(
-            int(psa_rotd_maximum_memory_allocation / (180 * sys.getsizeof(comp_0[0]))),
+            int(
+                psa_rotd_maximum_memory_allocation / (180 * sys.getsizeof(waveforms[0]))
```
lispandfound, line 303:
```suggestion
                psa_rotd_maximum_memory_allocation / (180 * len(w) * waveforms[0].nbytes)
```

tests/test_ims.py
```diff
@@ -115,15 +118,22 @@ def test_rotd_psa_values():
...
-    assert result.shape == (len(comp_0), len(w), 3)
-    assert result[0, 0, 2] == pytest.approx((2 * np.pi) ** 2 * np.sqrt(2), abs=1e-3)
+    w = 2 * np.pi * np.array([1.0], dtype=np.float32)
+
+    out = ims.rotd_psa_values(sample_waveforms, sample_dt, w, step=1)
+    rotd_psa = out.rotd_psa
+
+    print("!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!")
+    print("results:")
+    print(rotd_psa)
```
lispandfound: You should remove these print statements, and instead use the assert to show them only if they are interesting.

lispandfound (changes requested):
As discussed in person with Andrew, will invert the logic of this PR to preserve the public interface of `rotd_psa_values` and do the stepping in `pseudo_spectral_acceleration` instead.

AndrewRidden-Harper:
@lispandfound I checked the run times and the current (better) version with the loop in `pseudo_spectral_acceleration()` has about the same run time as the previous version with the loop in `rotd_psa_values()`.

joelridden (approved):
Nice work

lispandfound (approved):
LGTM! Nice work.

PR #186, opened 2025-07-16
fix: make obspy an optional dependency

lispandfound:
Makes `obspy` an optional dependency. Installed as `IM_calculation[ko_matrix]`


lispandfound:
The requirements.txt file was removed because it doesn't support optional dependencies or dependency groups.

joelridden (approved):

